PR #3, opened 2016-11-01
replcae cdiff with colordiff in automated test

diddledan:
`cdiff` is meant for interactive use, as you discovered. You have worked-around this by replacing the `less` binary on the system with a wrapper around `tail`. This PR removes that dirty hack in favour of using `colordiff` rather than the interactive wrapper `cdiff`.


jakubmisek:
Thanks!

bfistein:
Thanks a lot Dan!

PR #19, opened 2017-03-08
Uniform project format for VS2015

MaitreDede:
Reworked projects to target SDK 1.0.0-preview2, so code can be handled with Visual Studio 2015.
Also reworked projet references and dependencies to remove warnings


jakubmisek:
Fantastic!

jakubmisek:
Can we update src/tools/? some .rsp file has hardcoded refs to .dlls like mysql so we should add ref to mssql as well.

MaitreDede:
Is it a good idea to rework this part to remove hardcoded `rsp` files and generate them "on the fly" ?

Also for the "peach" tool to rely on reflection instead of a an hardcoded list of assemblies ? (maybe assembly references lookup)

jakubmisek:
@roberthusak can we generate rsp files on the fly ? or even better - use our roslyn API (PhpCompilation) and compile tests without rsp at all?

jakubmisek:
+1 for reflection in peach.exe ... smthng like `typeof(object).Assembly.Location` would be much better

roberthusak:
Yes, it should be possible to get rid of the hardcoded rsps - for example, we can use `dotnet restore` to a dummy project and then gather the resulting references to assemblies, similarily as in the language server. We'll need to rebuild the testing anyway in order not to parse all the references again with every *.php test file. So, I'll keep that in mind.

PR #23, opened 2017-03-12
phpinfo base implementation

MaitreDede:
Some bits for phpinfo implementation


jakubmisek:
Thanks! We'll merge the pull request and we'll make some changes
- there must be a non-html output for console apps as well
- the first overload of phpinfo() is redundant, compiler makes automatic conversion from int|long -> enum
- the tag writer may be replaced with XmlWriter or an abstraction that handles console and html outputs

jakubmisek:
Thanks for the implementation one more time. 

I've added some TODOs with API that provide extensions and their configurations.

Also we may rename HtmlTagWriter to a "PhpInfoWriter" with two implementations for HTML and plain Text (for CLI mode)

PR #35, opened 2017-04-14
Fix when Kestrel is listening on a unix socket

jdluzen:
This fixes a crash when behind e.g. nginx, and listening on a unix socket.


jakubmisek:
Thanks!

PR #47, opened 2017-05-16
Preserve conditioned expressions with yield statements

petrroll:


src/Peachpie.CodeAnalysis/Utilities/EnumeratorExtension.cs
```diff
@@ -193,5 +193,10 @@ public static ImmutableArray<T> AsImmutableSafe<T>(this IList<T> list)
                 return list.ToImmutableArray();
             }
         }
+
+        public static ImmutableArray<T> EncapsulateInImmutableArray<T>(this T item)
```
jakubmisek: remove <- ImmutableArray.Create<T>(T)

src/Peachpie.CodeAnalysis/Utilities/PhpOperationExtensions.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace Pchp.CodeAnalysis.Semantics
+{
+    public static class PhpOperationExtensions
+    {
+        #region ExpressionExtensions
+        /// <summary>
+        /// Returns whether the expression is safe to be evaluated multiple times i.e it has no side effects.
+        /// </summary>
+        public static bool IsSafeToEvalMultipleTimes(this BoundExpression expr)
```
jakubmisek: remove, nothing will be evaluated multiple times

src/Peachpie.CodeAnalysis/Semantics/PhpOperationVisitor.cs
```diff
@@ -252,6 +252,12 @@ public virtual void VisitYieldStatement(BoundYieldStatement boundYieldStatement)
             Accept(boundYieldStatement.YieldedKey);
         }
 
+        public virtual void VisitConditedStatement(BoundConditionedStatement x)
```
jakubmisek: remove, use ConditionalEdge

src/Peachpie.CodeAnalysis/Semantics/BoundStatement.cs
```diff
@@ -314,4 +314,38 @@ public override void Accept(OperationVisitor visitor)
         public override TResult Accept<TArgument, TResult>(OperationVisitor<TArgument, TResult> visitor, TArgument argument)
             => visitor.VisitReturnStatement(this, argument);
     }
+
+    /// <summary>
+    /// Represents a statement that is evaluated only if the condition is true.
+    /// </summary>
+    public partial class BoundConditionedStatement : BoundStatement, IIfStatement
```
jakubmisek: remove, use ConditionalEdge

src/Peachpie.CodeAnalysis/FlowAnalysis/ExpressionAnalysis.cs
```diff
@@ -1712,6 +1712,20 @@ public override void VisitConditional(BoundConditionalEx x)
             x.TypeRefMask = trueExpr.TypeRefMask | x.IfFalse.TypeRefMask;
         }
 
+        public override void VisitConditedStatement(BoundConditionedStatement x)
```
jakubmisek: remove, use ConditionalEdge

src/Peachpie.CodeAnalysis/CodeGen/Graph/BoundStatement.cs
```diff
@@ -367,4 +366,20 @@ private void setAsPhpValueOnGenerator(CodeGenerator cg, BoundExpression valueExp
             cg.EmitCall(ILOpCode.Call, setMethod);
         }
     }
+
+    partial class BoundConditionedStatement
```
jakubmisek: remove, we have ConditionalEdge

petrroll:
The problem is that I can't create a ConditionalEdge from SemanticsBinder. I'd somehow have to communicate the intend to make these statements conditioned(*) from SB to BuilderVisitor which would - IMO - be even worse than current solution.

(*) In a way that doesn't slow down / increase memory footprint for non-conditioned statements which constitute the majority of bound statements.

---

A way to do it would be to make BoundConditionedStatement only a shell (without .Emit, ...) and then loop trough all retrieved PreBoundStatements in BuilderVisitor and construct ConditionalEdge for found BoundConditionedStatement. That would, however, be way more work and frankly doesn't feel that clean either.

Also, there might be a problem with creating new edges inside certain constructs (didn't really think about that all that much, so can't say for sure).

src/Peachpie.CodeAnalysis/Utilities/PhpOperationExtensions.cs
petrroll: Yep, it's an old artifact. Removed.

src/Peachpie.CodeAnalysis/Utilities/EnumeratorExtension.cs
petrroll: Removed.

PR #49, opened 2017-06-08
Type and reachability analysis enhancement

roberthusak:
- Add type analysis tests.
- Use is_null(), is_int() etc. in type analysis.
- Enhance usage of comparison with null in type analysis.
- Use type analysis information to produce constant values of type checking expressions.
- Reflect constant conditions in reachability analysis.


src/Peachpie.CodeAnalysis/FlowAnalysis/ConditionBranch.cs
```diff
@@ -28,9 +28,22 @@ internal static class ConditionBranchEnum
         /// <summary>
         /// Switches <see cref="ConditionBranch.ToTrue"/> and <see cref="ConditionBranch.ToFalse"/>.
         /// </summary>
-        public static ConditionBranch NegativeBranch(ConditionBranch branch)
+        public static ConditionBranch NegativeBranch(this ConditionBranch branch)
         {
             return (ConditionBranch)(-((int)branch));
         }
+
+        public static bool? ToBool(this ConditionBranch branch)
```
jakubmisek: comment, what is a branch to boolean ?

src/Peachpie.CodeAnalysis/Semantics/PhpOperationVisitor.cs
```diff
@@ -130,7 +130,11 @@ public virtual void VisitList(BoundListEx x)
 
         public virtual void VisitFieldRef(BoundFieldRef x)
         {
-            VisitTypeRef(x.ParentType);
+            if (x.ParentType != null)
```
jakubmisek: move this `if` into VisitTypeRef

src/Peachpie.CodeAnalysis/Symbols/Source/SourceFileSymbol.cs
```diff
@@ -109,6 +109,24 @@ internal string DirectoryRelativePath
             }
         }
 
+        /// <summary>
+        /// Gets enumeration of all routines (global code, functions and methods) within the file.
+        /// </summary>
+        public IEnumerable<SourceRoutineSymbol> AllRoutines    // all functions + global code + methods + lambdas
```
jakubmisek: seams like a duplicity code

src/Peachpie.CodeAnalysis/Syntax/PhpSyntaxTree.cs
```diff
@@ -206,6 +206,21 @@ public override FileLinePositionSpan GetLineSpan(TextSpan span, CancellationToke
             return new FileLinePositionSpan(_source.FilePath, _source.LinePosition(span.Start), _source.LinePosition(span.End));
         }
 
+        /// <summary>
+        /// Returns the offset of the location specified by (zero-based) line and character from the start of the file.
+        /// In the case of invalid line, -1 is returned.
+        /// </summary>
+        public int GetPosition(LinePosition linePosition)
```
jakubmisek: there is already a method for that

src/Peachpie.CodeAnalysis/Utilities/ConstantValueExtensions.cs
```diff
@@ -46,6 +46,34 @@ public static bool IsNull(this Optional<object> value)
...
+        /// </summary>
+        public static bool EqualsOptional(this Optional<object> value, Optional<object> other)
+        {
+            if (value.HasValue != other.HasValue)
+            {
+                return false;
+            }
+            else if (!value.HasValue)
+            {
+                return true;
```
jakubmisek: Equals is doesn't have `value` ? comment

src/Peachpie.CodeAnalysis/FlowAnalysis/ConditionBranch.cs
roberthusak: Thanks for the review! Commented in 8a703772f93f0e141e66b944358489a5942fae23.

src/Peachpie.CodeAnalysis/Utilities/ConstantValueExtensions.cs
roberthusak: Commented in 8a703772f93f0e141e66b944358489a5942fae23.

src/Peachpie.CodeAnalysis/Symbols/Source/SourceFileSymbol.cs
roberthusak: OK, moved in 7c68d1a4be0c115e3eb3e79be7ff12a5405eac3f.

src/Peachpie.CodeAnalysis/Syntax/PhpSyntaxTree.cs
roberthusak: We can, but it is used from the language server, so it must be public. Should I make AstUtils public, then?

src/Peachpie.CodeAnalysis/Semantics/PhpOperationVisitor.cs
roberthusak: Well, I'll do it if you want, but I don't understand the reason. When inheriting a visitor, I was surprised that VisitTypeRef can be called with null.  The null check must be then done in the inherited visitor. Isn't it better to prevent possible null dereference and save an unnecessary method call by keeping the check in VisitFieldRef? Or am I missing something?

jakubmisek: Just an unnecessary piece code.
The method handles `null` already for the reason so any call to `VisitTypeRef` can be less complex, less of conditional code, less less braces etc. ... otherwise you could copy/paste `if (blahblah != null) { VisitTypeRef(blahblah); }` for 16 times and maintain it to the rest of our lives

src/Peachpie.CodeAnalysis/FlowAnalysis/ConditionBranch.cs
jakubmisek: ah, so it is not a conversion to a Boolean but rather its expected result.. maybe rename it to something else ? like `TargetValue` ?

src/Peachpie.CodeAnalysis/Semantics/PhpOperationVisitor.cs
roberthusak: Ah.. I see.. thanks for the explanation. Moved in d68f95462d7da012483a47ab2bf01842afaa6079.

src/Peachpie.CodeAnalysis/Syntax/PhpSyntaxTree.cs
roberthusak: OK, moved in 22b1016a632be4635aacb51ce4857c14f7e5bdde.

src/Peachpie.CodeAnalysis/FlowAnalysis/ConditionBranch.cs
roberthusak: OK, that sounds nice - 269dfcf393c008c38976c5c92d5785103e8bfc20.

PR #52, opened 2017-06-23
Fixes nested conditional expressions with yields.

petrroll:
- The CreateConditionalEdge in SB did not connect the conditioned blocks properly.


petrroll:
Do not merge yet, last commit does not work properly.

petrroll:
Last commit fixed with --force push. Should be ok now.
PR actually includes two fixes. See their respective tests for more details.

Issue #196, opened 2018-03-21
PHP ini configuration not implemented.


gordon-matt:
`ini_get('')`, `ini_set('', ...)` and other functions pertaining to PHP configuration are not implemented. In my case, the code is trying to **get** and/or **set** the following values:

`memory_limit`
`max_execution_time`
~~`post_max_size`~~
~~`allow_url_fopen`~~
~~`safe_mode`~~ // removed in PHP 5.4
~~`disable_functions`~~ // in https://github.com/peachpiecompiler/peachpie/commit/326aa52c09fd1dea46f939c1e7a250471b01c6d0
~~`upload_max_filesize`~~
~~`open_basedir`~~

The file manager I am trying to run needs these and is not generating thumbnails, at least in part due to `memory_limit` missing.

All of the above and more I found in Phalanger, here: [PhpIni.CLR.cs](https://github.com/DEVSENSE/Phalanger/blob/master/Source/ClassLibrary/PhpIni.CLR.cs).

Will this be a lot of effort?

**NOTE**: This is related to issue #185

gordon-matt:
OK I looked into the Peachpie code a bit deeper and I can see you have actually done some of the ini configuration. I just noticed `static PhpValue GsrCore(static PhpValue GsrCore(Context ctx, IPhpConfigurationService config, string option, PhpValue value, IniAction action)` in **Options.cs**... but it's missing some functions like `memory_limit`.

jakubmisek:
Good point. Few of them are already supported. Can you please make a list of options used within `ini_get` and `ini_set` separately? Because `ini_get` is easy to implement for all of them, `ini_set` might be tricky since it is mostly not possible in .NET (the configuration is somewhere else and we can't change it once the app starts)

jakubmisek:
`memory_limit` cannot be set in .NET, however, there is a bug in Responsive File Manager.
 They should check for `$memory_limit` is `-1` (https://github.com/gordon-matt/peachpie-responsive-file-manager/blob/3fbb7a7e8467e48baa9e6985def53b5d87b7597e/Website/filemanager/include/utils.php#L587).

jakubmisek:
`max_execution_time` will be implemented

gordon-matt:
[...]

PR #197, opened 2018-03-21
Added ini config values for `memory_limit` and `disable_functions`

gordon-matt:
Related to issue #196

### DIFF
--- a/src/Peachpie.Library/Options.cs
+++ b/src/Peachpie.Library/Options.cs
@@ -123,6 +123,11 @@ namespace Pchp.Library
                     Debug.Assert(action == IniAction.Get);
                     return (PhpValue)string.Empty;
 
+                case "disable_functions": return (PhpValue)string.Empty;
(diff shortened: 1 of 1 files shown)


src/Peachpie.Library/Options.cs
```diff
@@ -123,6 +123,11 @@ static PhpValue GsrCore(Context ctx, IPhpConfigurationService config, string opt
                     Debug.Assert(action == IniAction.Get);
                     return (PhpValue)string.Empty;
 
+                case "disable_functions": return (PhpValue)string.Empty;
+                case "memory_limit": return (PhpValue)(-1);
+                //TODO: case "safe_mode": return ???;
```
jakubmisek: this has been removed in PHP 5.4, but we can add it and return `false`

jakubmisek:
nice, thanks! we have merged these options and added several others as well

Issue #200, opened 2018-03-28
Missing functions for MultiByteString


gordon-matt:
I needed the following:

```
mb_http_output
mb_http_input
mb_language
mb_regex_encoding
mb_detect_encoding
```

I have implemented `mb_language`, `mb_regex_encoding` and a bunch of others that I didn't need, but that were missing and seemed easy to implement. I need help with `mb_http_output`, `mb_http_input` and `mb_detect_encoding`. A new PR is coming your way soon with the extra code. The methods I need help with are throwing **NotImplementedException**.

gordon-matt:
Just checked in a possible solution for `mb_http_input` and `mb_http_output` in the PR (Needs testing though)

gordon-matt:
I see this is now done. Thanks. Closing issue.

PR #201, opened 2018-03-28
Additional functions for MultiByteString

gordon-matt:
Please see issue #200 for further details

### DIFF
--- a/src/Peachpie.Library/posix.cs
+++ b/src/Peachpie.Library/posix.cs
@@ -74,7 +74,7 @@ namespace Pchp.Library
         /// <param name="useLimit">True if you want to use previous parameter.</param>
         /// <param name="ignoreCase">True if <c>str</c> is matched case insensitive, false otherwise.</param>
         /// <returns>Array containing parts of str</returns>
-        public static PhpArray DoSplit(Context ctx, string pattern, string str, int limit, bool useLimit, bool ignoreCase)
+        private static PhpArray DoSplit(Context ctx, string pattern, string str, int limit, bool useLimit, bool ignoreCase)
         {
             System.Array sAr;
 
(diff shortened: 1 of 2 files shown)


gordon-matt:
Sure, no worries - take your time.

bfistein:
Sorry about the delay @gordon-matt, but we will take a look at this PR momentarily. Thanks for understanding.

gordon-matt:
No problem @bfistein . I understand you guys are very busy.

src/Peachpie.Library/posix.cs
```diff
@@ -74,7 +74,7 @@ public static PhpArray spliti(Context ctx, string pattern, string str, int limit
         /// <param name="useLimit">True if you want to use previous parameter.</param>
         /// <param name="ignoreCase">True if <c>str</c> is matched case insensitive, false otherwise.</param>
         /// <returns>Array containing parts of str</returns>
-        private static PhpArray DoSplit(Context ctx, string pattern, string str, int limit, bool useLimit, bool ignoreCase)
+        public static PhpArray DoSplit(Context ctx, string pattern, string str, int limit, bool useLimit, bool ignoreCase)
```
jakubmisek: you cannot make a method public just like that - this would introduce a public PHP function `DoSplit`

gordon-matt: Ah, the reason that is changed to a `public` method is because I was working on some other function that required me to call that. I'm not 100% sure, but I think it was `mb_split`. You can see you guys do the exact same thing in Phalanger [here](https://github.com/DEVSENSE/Phalanger/search?utf8=%E2%9C%93&q=DoSplit&type=). Anyway, for some reason I later removed that function (decided not to implement it for some reason - maybe because I didn't need it). I can definitely change it back to a private method if you want, but maybe you should first see if you want to copy the `mb_string` function from Phalanger, which will need that method to be public. Let me know. Thanks.

jakubmisek: It should be put in another helper class or be marked with [PhpHidden] attribute; please just ensure there won't be a callable function `DoSplit` after the change.

gordon-matt: I have reverted to a private access modifier. All should be okay now.

gordon-matt:
@jakubmisek The change was made, as requested. Do I need to do something else?

jakubmisek:
@gordon-matt great, will merge, thank you

PR #209, opened 2018-05-23
Cross target libraries to .NETSTANDARD 2.0

iamcarbon:
This commit adds .NETSTANDARD2.0 targets to the peachpie libraries to reduce the dependencies on modern platforms. 

It also switches out the Oracle MySql library with the open source MySQLConnector library. This is used in the latest TechEmpower benchmarks, much faster than the Oracle library, and isn't subject to Oracles licensing terms.

### DIFF
--- a/src/PDO/Peachpie.Library.PDO.MySQL/Peachpie.Library.PDO.MySQL.csproj
+++ b/src/PDO/Peachpie.Library.PDO.MySQL/Peachpie.Library.PDO.MySQL.csproj
@@ -2,7 +2,7 @@
 
   <Import Project="..\..\..\build\Targets\Settings.props" />
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.6;netstandard2.0;net46</TargetFrameworks>
+    <TargetFrameworks>netstandard1.5;netstandard2.0;net46</TargetFrameworks>
     <PreserveCompilationContext>true</PreserveCompilationContext>
     <AssemblyName>Peachpie.Library.PDO.MySQL</AssemblyName>
     <PackageId>Peachpie.Library.PDO.MySQL</PackageId>
@@ -14,7 +14,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="MySql.Data" Version="7.0.6-IR31" />
+    <PackageReference Include="MySqlConnector" Version="0.40.4" />
   </ItemGroup>
 
 </Project>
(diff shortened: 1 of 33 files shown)


iamcarbon:
Not reproducing this test failure locally. Looking into this more.

screen shot 2018-05-23 at 2 49 08 pm

iamcarbon:
Ready for review / feedback.

iamcarbon:
Some of our dependencies are moving their minimum target to .NETSTANDARD 2.0. Should we also consider moving up our minimum target to .NETSTANDARD 2.0?

src/Peachpie.Library.MySql/MySqlConnectionResource.cs
```diff
@@ -48,7 +48,7 @@ public override void ClosePendingReader()
             var myreader = (MySqlDataReader)_pendingReader;
             if (myreader != null)
             {
-                myreader.Close();   // we have to call Close() on MySqlDataReader, it is declared as non-virtual!
```
jakubmisek: Is this safe now? In recent version of MySql connector, `Dispose()` didn't close the underlaying connection causing huge resources leaking and eventually applications to stop being able to open new connections.

iamcarbon: The underlying connection will be closed if we set CommandBehavior.CloseConnection flag when creating the command. Otherwise, the pooling behavior specified in the connection string should control the lifetime of the connections.

https://github.com/mysql-net/MySqlConnector/blob/master/src/MySqlConnector/MySql.Data.MySqlClient/MySqlDataReader.cs#L285

jakubmisek: good it seams they fixed this

src/PDO/Peachpie.Library.PDO.MySQL/Peachpie.Library.PDO.MySQL.csproj
```diff
@@ -14,12 +14,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="MySql.Data" Version="7.0.6-IR31" />
+    <PackageReference Include="MySqlConnector" Version="0.40.4" />
```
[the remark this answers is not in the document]
iamcarbon: It's an open source clean port of the MySql data library focused on performance without the bloat and licensing ambiguity of the Oracle provided library.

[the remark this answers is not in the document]
jakubmisek: nice

jakubmisek:
there is one test failing - `/tests/scripting/eval_001.php` - probably due to some references changed? we have to take a look on that

jakubmisek:
Thanks!

jakubmisek:
We have to revert back to Oracle MySQL Connector, several required connection options are not supported by the one you used (namely `allowzerodatetime`, `Interactive`, `Pipe`, `characterset`).

PR #213, opened 2018-06-06
Remove TypeInfo

iamcarbon:
Replaces TypeInfo with Type.

### DIFF
--- a/src/Peachpie.Runtime/Dynamic/Cache.cs
+++ b/src/Peachpie.Runtime/Dynamic/Cache.cs
@@ -74,12 +74,12 @@ namespace Pchp.Core.Dynamic
 
         public static class Properties
         {
-            public static readonly PropertyInfo PhpValue_Object = Types.PhpValue[0].GetTypeInfo().GetDeclaredProperty("Object");
-            public static readonly FieldInfo PhpValue_Void = Types.PhpValue[0].GetTypeInfo().GetDeclaredField("Void");
-            public static readonly FieldInfo PhpValue_Null = Types.PhpValue[0].GetTypeInfo().GetDeclaredField("Null");
-            public static readonly PropertyInfo PhpValue_False = Types.PhpValue[0].GetTypeInfo().GetDeclaredProperty("False");
-            public static readonly PropertyInfo PhpValue_True = Types.PhpValue[0].GetTypeInfo().GetDeclaredProperty("True");
-            public static readonly FieldInfo PhpNumber_Default = Types.PhpNumber[0].GetTypeInfo().GetDeclaredField("Default");
+            public static readonly PropertyInfo PhpValue_Object = Types.PhpValue[0].GetProperty("Object", BindingFlags.DeclaredOnly);
+            public static readonly FieldInfo PhpValue_Void = Types.PhpValue[0].GetField("Void", BindingFlags.DeclaredOnly);
+            public static readonly FieldInfo PhpValue_Null = Types.PhpValue[0].GetField("Null", BindingFlags.DeclaredOnly);
+            public static readonly PropertyInfo PhpValue_False = Types.PhpValue[0].GetProperty("False", BindingFlags.DeclaredOnly);
+            public static readonly PropertyInfo PhpValue_True = Types.PhpValue[0].GetProperty("True", BindingFlags.DeclaredOnly);
+            public static readonly FieldInfo PhpNumber_Default = Types.PhpNumber[0].GetField("Default", BindingFlags.DeclaredOnly);
         }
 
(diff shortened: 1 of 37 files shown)


src/Peachpie.Runtime/Reflection/TypeFields.cs
```diff
@@ -41,7 +41,7 @@ public class TypeFields
 
         internal TypeFields(Type tinfo)
         {
-            _fields = tinfo.GetFields(BindingFlags.Instance | BindingFlags.DeclaredOnly).Where(_IsAllowedField).ToDictionary(_FieldName, StringComparer.Ordinal);
+            _fields = tinfo.GetFields().Where(_IsAllowedField).ToDictionary(_FieldName, StringComparer.Ordinal);
```
jakubmisek: as I know, this has a different semantic - we MUST collect only declared fields (not fields from base type)

src/Peachpie.Runtime/Reflection/TypeFields.cs
```diff
@@ -56,10 +56,11 @@ internal TypeFields(Type tinfo)
                     staticscontainer = staticscontainer.MakeGenericType(tinfo.GenericTypeArguments);
                 }
 
-                _staticsFields = staticscontainer.GetFields(BindingFlags.Instance | BindingFlags.DeclaredOnly).ToDictionary(_FieldName, StringComparer.Ordinal);
+                _staticsFields = staticscontainer.GetFields().ToDictionary(_FieldName, StringComparer.Ordinal);
```
jakubmisek: as above

src/Peachpie.Runtime/Reflection/TypeFields.cs
```diff
@@ -56,10 +56,11 @@ internal TypeFields(Type tinfo)
...
             }
 
-            _properties = tinfo.GetProperties(BindingFlags.Instance | BindingFlags.DeclaredOnly).Where(_IsAllowedProperty).ToDictionary(_PropertyName, StringComparer.Ordinal);
+            _properties = tinfo.GetProperties().Where(_IsAllowedProperty).ToDictionary(_PropertyName, StringComparer.Ordinal);
```
jakubmisek: as above

src/Peachpie.Runtime/Reflection/PhpTypeInfo.cs
```diff
@@ -205,9 +199,9 @@ TObjectCreator BuildCreatorProtected()
                 {
                     var ctorsList = new List<ConstructorInfo>();
                     bool hasProtected = false;
-                    foreach (var c in _type.GetConstructors(BindingFlags.DeclaredOnly))
+                    foreach (var c in _type.GetConstructors(BindingFlags.Instance | BindingFlags.Public))
```
jakubmisek: this breaks the code - we need all no private ctors

src/Peachpie.Runtime/Reflection/PhpTypeInfo.cs
```diff
@@ -177,9 +171,9 @@ TObjectCreator BuildCreatorPrivate()
                 {
                     var ctorsList = new List<ConstructorInfo>();
                     bool hasPrivate = false;
-                    foreach (var c in _type.GetConstructors(BindingFlags.DeclaredOnly))
+                    foreach (var c in _type.GetConstructors(BindingFlags.Instance))
```
jakubmisek: + `DeclaredOnly` on current type only, not base types !!!

src/Peachpie.Runtime/Reflection/PhpTypeInfo.cs
jakubmisek: not base types, `DeclaredOnly` !!

src/Peachpie.Runtime/Dynamic/Cache.cs
```diff
@@ -123,46 +119,19 @@ public static class Object
...
-                {
-                    if (Enumerable.SequenceEqual(ptypes, ps.Select(p => p.ParameterType)))
-                    {
-                        return ctor;
-                    }
-                }
-            }
-
-            throw new ArgumentException();
+            return type.GetConstructor(ptypes) ?? throw new ArgumentException();
```
jakubmisek: `DeclaredOnly`

jakubmisek:
These commits change internal reflection significantly, used method calls have different semantics. I also think. there is no reason for these modifications.

iamcarbon:
Agreed. Backing out of this change.

jakubmisek:
I'm always little bit scared when so many magic code is being changed at once : )

PR #214, opened 2018-06-06
Improve Code Quality

iamcarbon:
- Removes the unneeded calls to GetTypeInfo() without changing any behaviors
- Remove and sort using statements
- Use expression bodies for simple properties
- Annotate a few readonly structs
- Use pattern matching
- Updates lang version to 7.2
- Updates travis image to dotnet v2.1

### DIFF
--- a/build/install_dependencies_ubuntu14.sh
+++ b/build/install_dependencies_ubuntu14.sh
@@ -1,14 +1,15 @@
 # Install .NET Core (see https://www.microsoft.com/net/core#ubuntu)
 
 # Add the dotnet apt-get feed (for .NET Core)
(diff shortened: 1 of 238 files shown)


iamcarbon:
Ready for review / feedback.

jakubmisek:
Jason, we appreciate the pull request, but I cannot merge your formatting changes. The PR makes too many (major) changes and too many unnecessary formatting changes.

The formatting changes in CodeAnalysis won't be (ever) accepted for two reasons - First you would be recognized as the author of these lines but you wouldn't be able to be responsible for them. Second, the code is kept in sync with Roslyn sources so we will be able to easily merge and maintain the code.

The change to .net core 2.1 has to be discussed and I don't see any reason to rely on newest .NET Core - we want to be compatible with as many other projects as possible.

`Annotate readonly structs` is fine

PR #215, opened 2018-06-09
Switch to MySqlConnector

bgrainger:
This repeats the switch to MySqlConnector originally made in #209, which was backed out in af1893fb9759818ecc0c0ae41ae40e479274331a due to incompatibility.

MySqlConnector 0.42.0 now supports the following connection string options, which were required by this code:
* `AllowZeroDateTime` mysql-net/MySqlConnector#507
* `Interactive` mysql-net/MySqlConnector#510
* `Pipe` mysql-net/MySqlConnector#454

~~I ran the TechEmpower Framework benchmarks locally and got the same number of requests per second before and after this change, so unfortunately it doesn't look like it'll bring an immediate performance improvements,~~ but it does [fix a lot of bugs](https://mysql-net.github.io/MySqlConnector/tutorials/migrating-from-connector-net/#fixed-bugs) in the ADO.NET driver.

_Edit:_ apparently the way I'm building the TFB benchmarks isn't bringing in MySqlConnector, which explains why performance was exactly the same...

### DIFF
--- a/src/PDO/Peachpie.Library.PDO.MySQL/Peachpie.Library.PDO.MySQL.csproj
+++ b/src/PDO/Peachpie.Library.PDO.MySQL/Peachpie.Library.PDO.MySQL.csproj
@@ -14,7 +14,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="MySql.Data" Version="8.0.11" />
+    <PackageReference Include="MySqlConnector" Version="0.42.0" />
   </ItemGroup>
 
 </Project>
--- a/src/Peachpie.Library.MySql/MySql.Functions.cs
+++ b/src/Peachpie.Library.MySql/MySql.Functions.cs
@@ -147,7 +147,7 @@ namespace Peachpie.Library.MySql
             var config = ctx.Configuration.Get<MySqlConfiguration>();
             Debug.Assert(config != null);
 
-            var connection_string = BuildConnectionString(config, ref server, username, password, (ConnectFlags)client_flags, characterset: "utf8mb4");
+            var connection_string = BuildConnectionString(config, ref server, username, password, (ConnectFlags)client_flags);
 
             bool success;
             var connection = MySqlConnectionManager.GetInstance(ctx)
@@ -179,7 +179,7 @@ namespace Peachpie.Library.MySql
             return mysql_connect(ctx, server, username, password, new_link, client_flags);
         }
 
-        internal static string BuildConnectionString(MySqlConfiguration config, ref string server, string user, string password, ConnectFlags flags, int connectiontimeout = 0, string characterset = null)
+        internal static string BuildConnectionString(MySqlConfiguration config, ref string server, string user, string password, ConnectFlags flags, int connectiontimeout = 0)
         {
             // connection strings:
             if (server == null && user == null && password == null && flags == ConnectFlags.None && !string.IsNullOrEmpty(config.ConnectionString))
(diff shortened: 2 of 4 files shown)


bgrainger:
If anyone has advice for running a local build of PeachPie with TFB, I'd love to hear it.

I forked FrameworkBenchmarks and added my fork of PeachPie as a submodule, then updated the test code to use a local build: https://github.com/bgrainger/FrameworkBenchmarks/compare/master...peachpie

The local PeachPie code I'm trying to use is: https://github.com/bgrainger/peachpie/compare/master...tfb2

When I try to run the benchmark, it fails at the `dotnet publish` step. If I leave `<Compile Include="**/*.php" />` in `Website.msbuildproj`, I get `peachpie: /usr/share/dotnet/sdk/2.1.300/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.DefaultItems.targets(245,5): error : Duplicate 'Compile' items were included. The .NET SDK includes 'Compile' items from your project directory by default. You can either remove these items from your project file, or set the 'EnableDefaultCompileItems' property to 'false' if you want to explicitly include them in your project file. For more information, see https://aka.ms/sdkimplicititems. The duplicate items were: 'dbraw.php'; 'fortune.php'; 'json.php'; 'plaintext.php'; 'updateraw.php' [/app/Website/Website.msbuildproj]`

If I remove that, I get: `peachpie: /usr/share/dotnet/sdk/2.1.300/Microsoft.Common.CurrentVersion.targets(2808,7): error MSB4057: The target "CreateManifestResourceNames" does not exist in the project. [/app/Website/Website.msbuildproj]`.

At this point I'm not sure how to proceed, so I'm unable to benchmark MySqlConnector.

jakubmisek:
looks good, thanks!

@roberthusak Can you help with running TechEmpower please?

@bgrainger We also deal with `ColumnFlags` so we would be able to get all the column properties without collecting the entire `GetColumnSchema` (expensive). Older version of Oracles MySQL connector did provide this API. (ability to call `GetResultSet` which is private now or something like `MySqlDataReader.GetFieldFlags` would be great advantage).
The performance of DB operations is very crucial and we would have a lot of benchmarking in future.

jakubmisek:
WordPress seems to run fine with the new connector, all the tests pass. Feels faster.

iamcarbon:
This is great!!! @bgrainger Thanks for making this happen!

roberthusak:
@bgrainger Looks nice, thanks! I hope to fix the problems with FrameworkBenchmarks in this PR: https://github.com/bgrainger/FrameworkBenchmarks/pull/1. Please let me know if it helps.

bgrainger:
Quick update: I have the TFB benchmarks running locally now, and the results are... not good. But after enabling logging, I can see that something is very wrong: every single connection that's returned to the pool is determined to be invalid and is destroyed. This basically completely negates the increased performance of connection pooling. I will investigate further.

bfistein:
@bgrainger thanks for your work on this, we really appreciate it.

bgrainger:
The performance problem is caused by some bad assumptions in MySqlConnector; namely, that most clients will set the initial database in the connection string and not change it. Of course, this isn't possible with PHP's `mysql_connect`/`mysql_select_db` API.

I've opened an issue to address this limitation: mysql-net/MySqlConnector#515

In the table below, I use `fix-515` to denote a hacked local copy of MySqlConnector I built that avoids the issue. But we can work around it with the current NuGet package by avoiding a call to `mysql_select_db` and instead inlining the database name in the SQL itself. This avoids a network roundtrip to change the DB (and have the server verify that it's valid, the user has permissions, etc.). Example code change: https://github.com/bgrainger/FrameworkBenchmarks/commit/a7748a12898f1dc8c3c761c6e67bdc31abfb5867

Another issue is "should we reset the connection or not?". I believe we should follow PHP MySQL API conventions here, but I'm not entirely sure what they are. I've opened #217 to discuss them. For now, I hard-coded `ConnectionReset=true/false` in the `MySqlConnection` connection string. Eventually, for the benchmarks, this should be moved to the appropriate PHP API. Note that `MySql.Data` has crippling performance issues when `ConnectionReset=true`, so they default it to `false` for performance. MySqlConnector defaults it to `true` for correctness, but you can opt out. I show `true` vs `false` in the table below.

Here are the performance numbers from the various combinations of options:

Library | Version | ConnectionReset | `mysql_select_db` | Req/s
--- | --- | --- | --- | ---
MySql.Data (previous) | 7.0.7-M61 | false (default) | yes | 6700
MySqlConnector | 0.42.0 | true (default) | yes | 2000
MySqlConnector | fix-515 | true (default) | yes | 7100
MySqlConnector | fix-515 | false | yes | 8200
MySqlConnector | 0.42.0 | true (default) | no | 8000
MySqlConnector | 0.42.0 | false | no | 9800

jakubmisek:
Nice!
- We can definitely merge your benchmarks changes to avoid `mysql_select_db`.
- Our PHP implementation of `mysql_select_db` can be smart enough to not re-select same DB in case the connection comes from the pool and has the DB already selected (is it possible?)
- `ConnectionReset` can be set to `false`, there wasn't any issue with it so far.

Issue #227, opened 2018-06-22
DBNAME option in datasource of MySQL PDO Driver


tobihille:
Hi,

sadly I can not create a connection to MySQL. The following sample code only echoes "FOO" and then dies.
<pre>
echo "FOO";
$dsn = 'mysql:dbname=dt;host=localhost';
$pdo = new PDO($dsn, "user", "password", []);
echo "BAR";
$result = $pdo->query('select * from admin_user');
echo "BAZ";
</pre>

When debugging in VS Code (by the way: this is amazing!) I see the following lines in the debugging-console after calling `new PDO(...)`:
<pre>
Exception thrown: 'System.ArgumentException' in MySqlConnector.dll
Exception thrown: 'System.ArgumentException' in System.Data.Common.dll
Exception thrown: 'System.ArgumentException' in System.Private.CoreLib.dll
Exception thrown: 'System.ArgumentException' in System.Private.CoreLib.dll
Exception thrown: 'System.ArgumentException' in System.Private.CoreLib.dll
</pre>

Just for reference: Magento uses (in my case) the following Data Source Name: `mysql:host=127.0.0.1;dbname=hackathon_m1;initStatements=SET NAMES utf8;model=mysql4;type=pdo_mysql;pdoType=;active=1`

I wish I could provide more data but every try (I stripped the $dsn - String down to the example mentioned in the example code) resulted in the same error.
Scince wordpress runs already on peachpie and does basically the same thing (https://github.com/WordPress/WordPress/search?q=new+PDO%28&unscoped_q=new+PDO%28) I assume at the moment I'm doing something wrong but I have no idea what this might be.

Kind regards
- Tobi

jakubmisek:
actually wordpress uses `mysqli` and `mysql` functions. It does not make use of `PDO`.

jakubmisek:
note: `PDO` is not tested at all, so we might try that first. Does it throw another exception? In VSCode, Debugging pane at the bottom, you can tick "All Exceptions" and it might give you detailed information about exceptions.

tobihille:
Thanks for the hint. By checking "All Exceptions" VS Code stopped with the following exception:
**CLR/System.ArgumentException**
<pre>
Exception thrown: 'System.ArgumentException' in MySqlConnector.dll: 'Option 'dbname' not supported.'
[...]

PR #228, opened 2018-06-23
Implemented a simple MySQL DSN wrapper. First step towards fixing #227

tobihille:
The current Version still has some issues, e.g. using "initstatements" still leads to an exception so it does not work for magento yet. I will try to fix this tomorrow if I can.

I assume it is not the nicest code because im quite unfamiliar with C# but even if you reject the PR you should get an idea what i wanted to fix, so I think it helps in both cases ;-)

### DIFF
--- a/src/PDO/Peachpie.Library.PDO/PDOMySQLWrapper.cs
+++ b/src/PDO/Peachpie.Library.PDO/PDOMySQLWrapper.cs
@@ -1,0 +1,26 @@
+﻿using System;
+
+namespace Peachpie.Library.PDO
+{
+    //TODO: Probably a similar class is needed for the other DBMS
+    class PDOMySQLWrapper
+    {
+        public static string convertDSN(string dsn)
+        {
+            if (dsn == null || dsn.Length == 0)
+            {
+                return dsn;
+            }
+
+            dsn = dsn.Replace("host", "Host");
+            dsn = dsn.Replace("port", "Port");
+            dsn = dsn.Replace("dbname", "Database");
+            //TODO: if unix_socket is provided the protocol probably needs to be set as well.
+            // but scince I have no way of testing this I do not touch it.
+            //dsn = dsn.Replace("unix_socket", "..."); 
+            dsn = dsn.Replace("charset", "CharSet");
+
+            return dsn;
+        }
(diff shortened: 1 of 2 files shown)


bfistein:
Your contributions are much appreciated @tobihille. We first have to fix our previous commit, and we'll review your PR very soon after. Thanks again.

jakubmisek:
something like that will be needed, ideally if `MySqlConnector` would understand `dbname` directly.

src/PDO/Peachpie.Library.PDO/PDOMySQLWrapper.cs
```diff
@@ -0,0 +1,26 @@
...
+                return dsn;
+            }
+
+            dsn = dsn.Replace("host", "Host");
+            dsn = dsn.Replace("port", "Port");
+            dsn = dsn.Replace("dbname", "Database");
+            //TODO: if unix_socket is provided the protocol probably needs to be set as well.
+            // but scince I have no way of testing this I do not touch it.
+            //dsn = dsn.Replace("unix_socket", "..."); 
+            dsn = dsn.Replace("charset", "CharSet");
```
jakubmisek: casing should not make difference

src/PDO/Peachpie.Library.PDO/PDOMySQLWrapper.cs
```diff
@@ -0,0 +1,26 @@
...
+        public static string convertDSN(string dsn)
+        {
+            if (dsn == null || dsn.Length == 0)
+            {
+                return dsn;
+            }
+
+            dsn = dsn.Replace("host", "Host");
+            dsn = dsn.Replace("port", "Port");
+            dsn = dsn.Replace("dbname", "Database");
```
jakubmisek: we have the idea, of course we cannot change "dbname" to "Database" in connection string, that would be an unpleasant sideeffect.

jakubmisek:
looking at the code, we should parse the DNS by ourselves in order to create valid connection options

jakubmisek:
Tobi, I'll close the PR and try to implement it more in general. Thanks, if you'd have any other hints please feel free to create another PR.

PR #261, opened 2018-08-09
PDO

jan-pavlovsky:
updates to PDO, corrected return types and argument types, exceptions, better parameterr binding...


DaveMagrathSmith:
Are there any plans to implement FETCH_UNIQUE?  (an associative fetch, where the resulting array is indexed by the value of the first column).  Would be nice, but I can work around if necessary.

Thanks,
Dave

PR #278, opened 2018-09-17
Deep function call stack fix

jan-pavlovsky:


jan-pavlovsky:
Ok. I was also thinking whether not to also implement the non-recursive version for SimpleVarUse? It is not necessary at the moment, but it could use the same alg and save us from some stackoverflow ex in the future.

PR #297, opened 2018-10-23
Add support for semantic transformations, refactor visitor hierarchy

roberthusak:
The semantic transformation phase is currently disabled, will be enabled once the ```Update``` method is implemented for bound expressions and used from the rewriters.


jakubmisek:
thanks, did some refactoring tho but looking forward to further semantic analyses :))

Q: Is the `BoundEdge.Source` really necessary? Cannot it be transformed the same as an IOperation ? (just returned a new one) ?

roberthusak:
Great, thanks :-)

Not really, it can be removed when the graph rewriter is modified to recreate the CFG instead of modifying it (that's gonna be in the next PR from this branch).

Issue #298, opened 2018-10-28
NullReferenceException during compilation


profMagija:
Trying to compile the following code generates a NullReferenceException:

```php
<?php

class C {
	const A = 1 | 1 << 1;
}
```

Stacktrace (from running a test):

```
Error Message:
 System.NullReferenceException : Object reference not set to an instance of an object.
Stack Trace:
   at Pchp.CodeAnalysis.CodeGen.CodeGenerator.IsLongOnly(TypeRefMask tmask) in /home/prof/RiderProjects/peachpie/src/Peachpie.CodeAnalysis/CodeGen/CodeGenerator.TypeRef.cs:line 33
   at Pchp.CodeAnalysis.Semantics.BoundBinaryEx.EmitBitOr(CodeGenerator cg, BoundExpression left, BoundExpression right) in /home/prof/RiderProjects/peachpie/src/Peachpie.CodeAnalysis/CodeGen/Graph/BoundExpression.cs:line 569
   at Pchp.CodeAnalysis.Semantics.BoundBinaryEx.Emit(CodeGenerator cg) in /home/prof/RiderProjects/peachpie/src/Peachpie.CodeAnalysis/CodeGen/Graph/BoundExpression.cs:line 113
   at Pchp.CodeAnalysis.CodeGen.CodeGenerator.EmitConvert(BoundExpression expr, TypeSymbol to) in /home/prof/RiderProjects/peachpie/src/Peachpie.CodeAnalysis/CodeGen/CodeGenerator.Convert.cs:line 1253
   at Pchp.CodeAnalysis.Symbols.SourceFieldSymbol.Pchp.CodeAnalysis.Symbols.IPhpPropertySymbol.EmitInit(CodeGenerator cg) in /home/prof/RiderProjects/peachpie/src/Peachpie.CodeAnalysis/CodeGen/Symbols/SourceFieldSymbol.cs:line 33
   at Pchp.CodeAnalysis.Symbols.SynthesizedStaticFieldsHolder.<>c__DisplayClass0_0.<EmitCtors>b__0(ILBuilder il) in /home/prof/RiderProjects/peachpie/src/Peachpie.CodeAnalysis/CodeGen/Symbols/SynthesizedStaticFieldsHolder.cs:line 45
[...]

PR #299, opened 2018-10-28
fix NullReferenceException in CodeGenerator.TypeRef

profMagija:
fixes issue #298


jakubmisek:
thanks! it has been fixed in a different way in https://github.com/peachpiecompiler/peachpie/commit/621ecd679396032bc021f1d3a667349dbead50ca so it takes the flow analysis into account as well.

PR #314, opened 2018-11-25
Enable semantic transformations

roberthusak:
- Implement ```Update``` on all the bound expressions, statements, blocks, edges and CFG.
- Refactor ```TransformationRewriter``` to use ```Update``` in an immutable fashion.
- Enable transformations in Release - repeat analysis and transform in cycle.
- Fix various bugs and increase the readability of the algorithm.


jakubmisek:
superb, merging

Issue #304, opened 2018-11-05
eval() causes request to hang on asp.net/net461


dlob:
When requesting the following sample php file (compiled with peachpie 0.9.12) with `Peachpie.RequestHandler.RequestHandler` on target framework `net461` it hangs and doesn't return anything:

```php
<?php
    eval("echo \"test\";");
?>
```

jakubmisek:
Would you have more details? It probably fails by .NET runtime with an assembly load exception. If you'd enable some kind of detailed logging of the ASP.NET app on your IIS/IIS Express.

dlob:
There isn't an exception. The request never returns any content. Since IISExpress musst be killed after a request, I assume there is something like an endless loop.

jakubmisek:
Thank you! We were not able to find this out.

jakubmisek:
If the PR #315 is confirmed, closing the issue. Thank you @dlob for finding out.

PR #315, opened 2018-11-27
Resolves deadlock in Peachpie.CodeAnalysis

dlob:
Fixes issue #304.

Explanation: https://stackoverflow.com/questions/14526377/why-does-this-async-action-hang


jakubmisek:
Thank you! Amazing work.

PR #325, opened 2018-12-03
PDO Statement - Get Column Meta Fix

jondmcelroy:
Beofre this fix, the column names were in time added to the PDOStatement object when a fetch was called. The problem is when getColumnMeta is called before a fetch occurs. The array keeping track of the column names is empty and a null reference exception for referencing a null array.

The solution was to combine the logic for initializing the column names and also call it in the other locations.


jakubmisek:
nice! thanks

PR #327, opened 2018-12-03
Objects - is_subclass_of works properly with same parent class

jondmcelroy:
The bug is the is_subclass returns true when you pass the same class as the parent and the child. It should return false.


jakubmisek:
thanks, i've also noticed there are some misssing parameters and `is_a` function not handling interfaces. Just adding.

PR #362, opened 2019-02-16
Fix bin operators binding for non-PhpValue operands.

### LINKED ISSUE #361: EmitAdd on void and smth.
```
<?php
function a(){
}

function b(){
	return 3;
}
echo(a() + b());
```

Expected: 3
Actual: Exception

petrroll:
Fixes #361


jakubmisek:
Nice work @petrroll !

PR #365, opened 2019-03-01
Enabled sourcelink & symbols

petrsvihlik:
...to enable easier debugging.

Docs:
- https://github.com/dotnet/sourcelink
- https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg


jakubmisek:
thanks, good idea.

Please put the sourcelink code into `\build\Targets\Settings.props` instead of every single project file.

petrsvihlik:
done

PR #375, opened 2019-03-13
Implement more diagnostics

roberthusak:
- Redundant cast
- Array key type and duplicity
- Cloning of non-objects
- Enumeration of non-iterables
- Class name case
- ```printf``` argument count with regards to the format string
- Missing ```parent::__construct``` call
- PHPDoc param and field type mismatch
- Unreachable code for ```===```, ```!==``` and ```instanceof```
- Missing static method call


src/Peachpie.CodeAnalysis/FlowAnalysis/Passes/DiagnosticWalker.cs
```diff
@@ -103,13 +109,56 @@ protected override void VisitCFGInternal(ControlFlowGraph x)
...
+                        var hintType = param.Type;
+                        TypeSymbol docType = null;
+
+                        var tmask = PHPDoc.GetTypeMask(TypeCtx, param.PHPDocOpt.TypeNamesArray, _routine.GetNamingContext());
+                        if (!tmask.IsVoid && !tmask.IsAnyType)
+                        {
+                            docType = DeclaringCompilation.GetTypeFromTypeRef(TypeCtx, tmask);
+                        }
+
+                        if (!docType?.IsOfType(hintType) ?? false)
```
jakubmisek: isn't it easier just to check `!docType.IsOfType(hintType)` right on line 149 :) All these `Nullable` and `??` and false ... is just an overkill

src/Peachpie.CodeAnalysis/FlowAnalysis/AnalysisFacts.cs
```diff
@@ -522,5 +523,52 @@ private static bool TryGetVariableHandle(IVariableReference boundvar, FlowState
...
+            return false;
+        }
+
+        /// <summary>
+        /// If present, transforms the given constant value to a string corresponding to the key under which the item is stored in an array.
+        /// </summary>
+        /// <param name="keyConst">Constant value of the key.</param>
+        /// <param name="keyString">If <paramref name="keyConst"/> contains a value, the key as a string.</param>
+        /// <returns>Whether the value was constant at all.</returns>
+        public static bool TryGetCanonicKeyStringConstant(Optional<object> keyConst, out string keyString)
```
jakubmisek: this causes a lot of string allocations :( 

the conversion to array key is described in runtime by struct `IntStringKey`. Can we do the similar?

return Tuple<string, long>
convert all numbers to long
convert string to long if possible, otherwise keep the string
else return false

this would avoid repetitious conversions of number -> string (which is the most common case) and would speedup the duplicity check

src/Peachpie.CodeAnalysis/FlowAnalysis/ExpressionAnalysis.cs
```diff
@@ -1551,7 +1584,15 @@ public override T VisitStaticFunctionCall(BoundStaticFunctionCall x)
 
                 candidates = Construct(candidates, x);
 
-                x.TargetMethod = new OverloadsList(candidates).Resolve(this.TypeCtx, x.ArgumentsInSourceOrder, VisibilityScope);
+                var method = new OverloadsList(candidates).Resolve(this.TypeCtx, x.ArgumentsInSourceOrder, VisibilityScope);
+                if ((method is MissingMethodSymbol || method is InaccessibleMethodSymbol)
+                    && type.LookupMember<IMethodSymbol>(Name.SpecialMethodNames.CallStatic.Value) != null)
+                {
+                    // __callStatic at runtime solves both inaccessible and missing method problems
+                    method = null;
```
jakubmisek: nice!
please add note like ~ `// TODO: remember and emit call to __callstatic directly (CallStaticMethodSymbol?)`

src/Peachpie.CodeAnalysis/FlowAnalysis/Passes/DiagnosticWalker.cs
roberthusak: It is :-) I just didn't realize that ```GetTypeFromTypeRef``` cannot return ```null```. Fixed in 2eee6f3ae66e8a265d3704329c9789ab05bc34bd.

src/Peachpie.CodeAnalysis/FlowAnalysis/AnalysisFacts.cs
roberthusak: Good idea, fixed in 2eee6f3ae66e8a265d3704329c9789ab05bc34bd.

src/Peachpie.CodeAnalysis/FlowAnalysis/ExpressionAnalysis.cs
roberthusak: OK, added in 2eee6f3ae66e8a265d3704329c9789ab05bc34bd.

Issue #345, opened 2019-01-10
Array value changed to reference after updating it through a reference iterator


crbrz:
```php
$a = [0,0,0];

foreach($a as &$value) { }    // increases RefCount for values, does not decrease it back

$b = $a;
$b[0] = 1; // it changes $a[0] as well (because it is &$a[0])

print_r( $a ); // expected: [0,0,0]; actual [1,0,0]
```

The values in `$a` shouldn't be affected by the changes made in `$b`. 

It happens with 0.9.28.

PR #382, opened 2019-03-20
Restore values changed to aliases by iterator (OrderedDictionary.Enumerator.CurrentValueAliased)

crbrz:
Fix for #345

### DIFF
--- a/src/Peachpie.Runtime/OrderedDictionary.cs
+++ b/src/Peachpie.Runtime/OrderedDictionary.cs
@@ -1580,6 +1580,8 @@ namespace Pchp.Core
         {
             FastEnumerator/*!*/_enumerator;
 
+            private HashSet<int> _aliasValueIndexes; // indexes where the value was changed to an alias
+
             internal OrderedDictionary UnderlayingArray => _enumerator.UnderlayingArray;
 
             public Enumerator(OrderedDictionary/*<TValue>*/ array)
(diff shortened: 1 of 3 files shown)


jakubmisek:
sadly, this is extremely performance+memory critical implementation,
`Dispose` is not guaranteed to be called,
`PhpAlias` already has `ReferenceCount` for these purposes ready.

This would have to be tightly integrated into all APIs and compiler in order to work properly

crbrz:
It's more of a workaround, each `foreach` loop that uses a reference variable to access the values exposes the array to unwanted changes (values in array are changed to `PhpAlias` when `Enumerator.CurrentValueAliased` property is accessed even if no changes are made to the value) which can lead to weird runtime bugs. Basically, with current implementation the PHP sources must be patched to avoid any `foreach` loops with reference variables for current value access.

From what I understand to change a value directly in the array, the code generated for `foreach` loops should use something like a property on the enumerator that returns a `ref PhpValue` to the current value in the array data but this means the compiler must know how to use that enumerator property in other expressions where the current value variable might show up.

jakubmisek:
It is more than that. The workaround introduces set of new issues. E.g. If you'd use any of variables inside the `foreach` loop by reference it would change them back to values.

The issue #345 is a sad side effect of PHP's reference counting, one quick fix always introduces another issue. Important to note; the issue is not about `foreach` at all - it is about reference counting in general. The `foreach` just makes use of it.

crbrz:
Even if no other variables are created by reference from `foreach` loop, the variable used to access the values by reference lives until the function/script finishes (I completely forgot about this).

I guess it's patching PHP sources where necessary until `PhpAlias` gets reference counting implemented. Btw, I noticed the changes in this commit [WORKAROUND: avoid aliasing $endpoints elements](https://github.com/iolevel/wpdotnet-sdk/commit/c030eb97552b7598de56129edf2fbbac7a72667d) were overwritten when WordPress was updated.

bfistein:
You're right @crbrz, happened in a previous PR and we didn't notice. Thanks a lot for your contributions!

jakubmisek:
fixed in compiler in https://github.com/peachpiecompiler/peachpie/commit/56f591c9ef733dd4f5865e6528bbe308ad726cfa

PR #408, opened 2019-04-05
fix for msbuild (Visual Studio) not logging the actual compiler excep…

crbrz:
…tions


jakubmisek:
thanks! (there is one failed test but not related to this)

PR #410, opened 2019-04-05
Implemented setTimestamp for DateTime and DateTimeImmutable and some PDO fixes

calvinbaart:
- PDOStatement::bindValue now uses the correct index for Positional arguments (1-indexed)
- PDOStatement::fetchColum and PDOStatement::fetchObject both fetch the next row now before retrieving data.

### DIFF
--- a/src/PDO/Peachpie.Library.PDO/PDOStatement.cs
+++ b/src/PDO/Peachpie.Library.PDO/PDOStatement.cs
@@ -523,7 +523,7 @@ namespace Peachpie.Library.PDO
                     m_pdo.HandleError(new PDOException("Supplied parameter index must be an integer."));
                     return false;
                 }
-                int paramIndex = (int)parameter;
+                int paramIndex = ((int)parameter) - 1;
 
                 if (paramIndex < m_positionalPlaceholders.Count)
                 {
@@ -1082,6 +1082,12 @@ namespace Peachpie.Library.PDO
                 return PhpValue.False;
             }
 
+            // read next row
+            if (!this.m_dr.Read())
+            {
+                return PhpValue.False;
+            }
+
             return this.ReadArray(false, true)[column_number].GetValue();
         }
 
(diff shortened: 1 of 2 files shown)


src/PDO/Peachpie.Library.PDO/PDOStatement.cs
```diff
@@ -1082,6 +1082,12 @@ public PhpValue fetchColumn(int column_number = 0)
                 return PhpValue.False;
             }
 
+            // read next row
+            if (!this.m_dr.Read())
```
jakubmisek: are we sure we want to read **next** row for every `fetchColumn` ?

calvinbaart: According to the PHP documentation it should fetch the next row:
https://www.php.net/manual/en/pdostatement.fetchcolumn.php

`PDOStatement::fetchColumn — Returns a single column from the next row of a result set`
`Returns a single column from the next row of a result set or FALSE if there are no more rows.`

So I interpreted that as it will always fetch a new row. I'll try to do a test in PHP to see how it behaves to confirm

jakubmisek: thank you, I was confused a little

calvinbaart: Just to be safe I ran the following test in PHP:
```php
$statement = $pdo->prepare("SELECT * FROM tasks");
$statement->execute();

$tmp1 = $statement->fetchColumn(1);
$tmp2 = $statement->fetchColumn(1);
$tmp3 = $statement->fetchColumn(1);

echo $tmp1 . "\r\n";
echo $tmp2 . "\r\n";
echo $tmp3 . "\r\n";
```

Where the tasks table is filled like this:
```
  id  |  name  |
  1   |  row1  |
  2   |  row2  |
  3   |  row3  |
```

The output in PHP is:
```
row1
row2
row3
```

jakubmisek: thanks, I've checked the sample in PHP's manual as well and you're right

bfistein:
@calvinbaart are you on Twitter? If you are and agree, we'd love to give you a shout out publicly for your great contributions.

calvinbaart:
@bfistein Yea, my twitter handle is @calvin_baart. Turns out I never used it for anything though

Issue #406, opened 2019-04-04
System.NotImplementedException: 'Cannot implicitly convert 'String' to 'String'.


crbrz:
Compiling this [project](https://github.com/crbrz/wpdotnet-sdk) with this command `dotnet build -c Debug -m:1 -nr:false` gives this error: `Cannot implicitly convert 'String' to 'String'. not implemented at C:\Work\wpdotnet-sdk\PeachPied.WordPress.StorefrontTheme\wp-content\themes\storefront\inc\storefront-functions.php(154, 1)`. Peachpie version is 0.9.38 with dotnet 2.2.202. The reported line number seems to be off by 1 (zero based?).

I've been getting these errors in a similar project but only at random times most likely depending on how the individual projects are distributed between the dotnet processes spawned to handle the solution build. I think the issue might have something to do with running Peachpie compiler directly inside dotnet (or msbuild) host process ([8b311e028f29cbf0b8f135689edfd73170da9820](https://github.com/peachpiecompiler/peachpie/commit/8b311e028f29cbf0b8f135689edfd73170da9820)). By forcing the build to run in a single dotnet process the error occurs each time the solution is built. The error doesn't occur when compiling with 0.9.36.

jakubmisek:
I see! Since the compiler lives in the MSBuild Task now and the process is reused by Visual Studio for subsequent compilations, all the static variables survive.. clearing the statics (https://github.com/peachpiecompiler/peachpie/pull/411) works for now, we'll try to fix it properly and make use of the MSBuild process pooling.

PR #411, opened 2019-04-06
reset resolved types from a previous compilation

crbrz:
Fix for #406


jakubmisek:
thanks for the fix, it helped us to find the cause of the issue. This has to be fixed by making `BoundTypeRefFactory` an instance on `PhpCompilation` to allow compilations in parallel within a single process and better reuse of resolved symbols.

jakubmisek:
fixed with https://github.com/peachpiecompiler/peachpie/commit/7ca861aca605aa8ec9c763963d718b32df4fbf51

thanks for the hint!

crbrz:
I don't think msbuild/dotnet builds two different projects in a process at the same time but having `BoundTypeRefFactory` as an instance in `PhpCompilation` is definitely better.

bfistein:
@crbrz are you on Twitter? If you are and agree, we'd love to give you a shout out publicly for your great contributions.

crbrz:
No, I'm not using Twitter.

Issue #413, opened 2019-04-07
Multiple set-cookie headers overwrite each other


calvinbaart:
Hi,

Laravel calls the header function with set-cookie for each cookie. The way PeachPie handles this causes it to only pick-up the latest set-cookie.

```php
header("set-cookie: test1=a;");
header("set-cookie: test2=b;");
header("set-cookie: test3=c;");
```

In the above code only the test3 cookie is send back. PHP sends them all back.

jakubmisek:
fixed with https://github.com/peachpiecompiler/peachpie/pull/414

thanks!

PR #414, opened 2019-04-07
Fixed #412 and #413

### LINKED ISSUE #412: Cloning DateTime doesn't copy over the modified Date
Hi,

Laravel uses a construction like this to set the cookie expiration dates:
```php
$date = (new DateTime())->modify("+120 minutes");
$cookieExpiration = clone $date;

if ($date->getTimestamp() === $cookieExpiration->getTimestamp()) {
    echo "equal";
} else {
    echo "not equal";
}
```

On PHP this returns `equal`. On PeachPie this returns `not equal`. In PeachPie the default constructor of DateTime is called when cloning causing it to be set to the current time.

calvinbaart:
- Implemented IPhpCloneable for DateTime and DateTimeImmutable.
- Implemented modify for DateTimeImmutable.
- Added tests for Cloning DateTime and DateTimeImmutable
- Multiple set-cookie headers are now concatenated together.

Fixes #412 and #413

### DIFF
--- a/src/Peachpie.AspNetCore.Web/RequestContextCore.cs
+++ b/src/Peachpie.AspNetCore.Web/RequestContextCore.cs
@@ -40,13 +40,15 @@ namespace Peachpie.AspNetCore.Web
             }
 
             StringValues newitem = new StringValues(value);
-            //StringValues olditem;
-            //if (_httpctx.Response.Headers.TryGetValue(name, out olditem))
-            //{
-            //    newitem = StringValues.Concat(olditem, newitem);
-            //}
+            if (name.EqualsOrdinalIgnoreCase("set-cookie"))
+            {
+                StringValues olditem;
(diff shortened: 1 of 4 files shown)


src/Peachpie.AspNetCore.Web/RequestContextCore.cs
```diff
@@ -40,13 +40,15 @@ void IHttpPhpContext.SetHeader(string name, string value)
             }
 
             StringValues newitem = new StringValues(value);
-            //StringValues olditem;
-            //if (_httpctx.Response.Headers.TryGetValue(name, out olditem))
-            //{
-            //    newitem = StringValues.Concat(olditem, newitem);
-            //}
+            if (name.EqualsOrdinalIgnoreCase("set-cookie"))
```
calvinbaart: I wasn't sure why the original code was commented here so I only enabled it for set-cookie now.

PR #423, opened 2019-04-11
Make sure the currently loaded assemblies get referenced during eval compilation

calvinbaart:


jakubmisek:
> It seams the PR contains some other changes as well.

All the assembly references for `eval'd` code should be resolved at https://github.com/peachpiecompiler/peachpie/blob/3e2e6861e66641b66157f45db8affba1928fce6b/src/Peachpie.Library.Scripting/PhpCompilationFactory.cs#L68Also 

In this way they get parsed and processed just once. The PR would introduce significant memory and performance overhead for any eval. Also I wouldn't recommend to list every referenced assembly of current AppDomain. This would include all the Asp.Net libraries, MVC and more, which is not desirable.

PR #425, opened 2019-04-14
Implemented ReflectionFunctionAbstract::hasReturnType and getReturnType

### LINKED ISSUE #419: Reflection has unimplemented methods required by PHPUnit and Mockery
The following methods need to be implemented for PHPUnit and Mockery:
- [x] ReflectionClass::implementsInterface
- [x] ReflectionParameter::isVariadic 
- [x] ReflectionMethod::hasReturnType
- [x] ReflectionMethod::getReturnType
- [x] ReflectionMethod::isProtected
- [x] ReflectionMethod::isPrivate

calvinbaart:
Fixes #419

### DIFF
--- a/src/Peachpie.Library/Reflection/ReflectionType.cs
+++ b/src/Peachpie.Library/Reflection/ReflectionType.cs
@@ -11,11 +11,16 @@ namespace Pchp.Library.Reflection
     [PhpType(PhpTypeAttribute.InheritName), PhpExtension(ReflectionUtils.ExtensionName)]
     public class ReflectionType
     {
-        public virtual bool allowsNull() { throw new NotImplementedException(); }
+        public string Name { get; set; }
+        public bool AllowsNull { get; set; }
+        public bool BuiltIn { get; set; }
 
-        public virtual bool isBuiltin() { throw new NotImplementedException(); }
 
-        public virtual string __toString() { throw new NotImplementedException(); }
+        public virtual bool allowsNull() => AllowsNull;
+
+        public virtual bool isBuiltin() => BuiltIn;
+
+        public virtual string __toString() => Name;
 
         public override string ToString() => __toString();
     }
@@ -23,7 +28,7 @@ namespace Pchp.Library.Reflection
     [PhpType(PhpTypeAttribute.InheritName), PhpExtension(ReflectionUtils.ExtensionName)]
     public class ReflectionNamedType : ReflectionType
     {
-        public virtual string getName() { throw new NotImplementedException(); }
+        public virtual string getName() => Name;
     }
 
 }
--- a/src/Peachpie.Runtime/Attributes.cs
+++ b/src/Peachpie.Runtime/Attributes.cs
@@ -288,6 +288,26 @@ namespace Pchp.Core
     }
 
     /// <summary>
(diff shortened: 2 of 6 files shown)


jakubmisek:
I was thinking if we can implement it just using .NET reflection without an explicit `[ReturnTypeAttribute]`

In this way it won't work for regular C# methods.

calvinbaart:
The problems I have encountered with that are that sometimes the optimizer will give the class a return type. For example in the following sample:
```php
function func1()
{
    return 10;
}
```
PHP will return false on hasReturnType but the compiled version will be
```csharp
long func1()
{
    return 10;
}
```

src/Peachpie.Library/Reflection/ReflectionType.cs
```diff
@@ -11,19 +11,24 @@ namespace Pchp.Library.Reflection
     [PhpType(PhpTypeAttribute.InheritName), PhpExtension(ReflectionUtils.ExtensionName)]
     public class ReflectionType
     {
-        public virtual bool allowsNull() { throw new NotImplementedException(); }
+        public string Name { get; set; }
+        public bool AllowsNull { get; set; }
+        public bool BuiltIn { get; set; }
```
jakubmisek: The class should not have properties `Name`, `AllowsNull` and `BuiltIn` visible (https://www.php.net/manual/en/class.reflectiontype.php)

They should be marked `internal` or `internal protected`.

src/Peachpie.Runtime/Attributes.cs
```diff
@@ -287,6 +287,26 @@ public sealed class NotNullAttribute : Attribute
...
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method)]
+    public sealed class ReturnTypeAttribute : Attribute
+    {
+        /// <summary>
+        /// The type containing exported functions.
+        /// </summary>
+        public string SyntaxType { get; private set; }
+
+        public bool AllowsNull { get; private set; }
```
jakubmisek: We already have attribute `[return: NotNull]` which should be emitted if the return type cannot be `null`.

src/Peachpie.Runtime/Attributes.cs
```diff
@@ -287,6 +287,26 @@ public sealed class NotNullAttribute : Attribute
 
     }
 
+    /// <summary>
+    /// Compiler generated attribute denoting that the method returns a specific type
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method)]
+    public sealed class ReturnTypeAttribute : Attribute
```
jakubmisek: I think this is redundant - we could rely on .NET type. Annotating every method with an attribute and using too much strings may not worth it.

jakubmisek:
Yes, it would annotate return types for private, static and global functions, maybe more in future. But is it problem for PHP apps?

jakubmisek:
Also nullable value types would be seen as `mixed` (because now they compile as `PhpValue` instead of `Nullable<long>` for example. :)

jakubmisek:
proposing CLR based ReflectionNamedType
https://github.com/peachpiecompiler/peachpie/commit/13f66d5bef0764cca77ba8bcf952064c39ed41db

This will work in general while we will fix/implement additional cases by adding custom attributes, when needed.

PR #450, opened 2019-06-20
Add RW lock to ScriptingProvider to prevent race conditions

roberthusak:
Add Read-Write locking to ```ScriptingProvider``` to prevent race conditions. It's not ensured that the same script won't be inserted multiple times in the particular list, because I expected it to be a corner case not important enough to waste resources to prevent it. But better check it please :-)


jakubmisek:
does not seem like a correct rw lock pattern, there is still the race condition. Let me check

PR #470, opened 2019-08-16
Reference the entry assembly when using eval. Fixes #422

### LINKED ISSUE #422: Mockery: eval doesn't autoload classes
Suppose we have a test.php with these contents:
```php
<?php

class TestClass
{
}
```

When executing the following code PeachPie will throw an exception saying TestClass can't be found:
```php
spl_autoload_register(function ($class) {
    require "test.php";
});

eval("new TestClass();");
```

On PHP this works as expected (program ends without output or exception)

calvinbaart:
Hopefully this is a better change then the previous PR in https://github.com/peachpiecompiler/peachpie/pull/423

When running eval the code doesn't reference assemblies linked by the entry assembly. This causes it to get issues when there is PHP code in another assembly. An example of this bug can be found here: https://github.com/calvinbaart/bug-peachpie-eval

Running assembly2 causes it to crash with the following error:

```
Unhandled Exception: System.InvalidOperationException: Type name 'TestClass' could not be resolved.
   at Pchp.Core.DefaultErrorHandler.Throw(PhpError error, String message)
   at Peachpie.Library.Scripting.Script.<>c__DisplayClass15_0.<CreateInvalid>b__0(Context ctx, PhpArray locals, Object this, RuntimeTypeHandle self)
   at Peachpie.Library.Scripting.Script.Evaluate(Context ctx, PhpArray locals, Object this, RuntimeTypeHandle self)
   at Pchp.Core.Operators.Eval(Context ctx, PhpArray locals, Object this, RuntimeTypeHandle self, String code, String currentpath, Int32 line, Int32 column)
   at <Root>.test_php.<Main>(Context <ctx>, PhpArray <locals>, Object this, RuntimeTypeHandle <self>) in /Users/calvinbaart/Development/peachpie-bug-sample/assembly2/test.php:line 7
   at <Script>.Main(String[] args)
```

### DIFF
--- a/src/Peachpie.Library.Scripting/PhpCompilationFactory.cs
+++ b/src/Peachpie.Library.Scripting/PhpCompilationFactory.cs
@@ -85,7 +85,12 @@ namespace Peachpie.Library.Scripting
             var list = types.Distinct().Select(ass => ass.GetTypeInfo().Assembly).ToList();
             var set = new HashSet<Assembly>(list);
 
-            list.Add(System.Reflection.Assembly.GetEntryAssembly());
+            // Add entry assembly if its loadable
+            var entryAssembly = System.Reflection.Assembly.GetEntryAssembly();
+
+            if (!entryAssembly.IsDynamic && entryAssembly.Location.Length != 0) {
+                list.Add(entryAssembly);
+            }
 
             for (int i = 0; i < list.Count; i++)
             {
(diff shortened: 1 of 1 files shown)


calvinbaart:
In this build: https://travis-ci.com/calvinbaart/laravel-peachpie-sample/builds/123644740 all the
``System.InvalidOperationException: Class 'Mockery\HigherOrderMessage' not found`` will disappear with this fix.

jakubmisek:
thank you for the PR @calvinbaart - we have to take a closer look on this. 

`eval'd` code should see the assembly itself (but `GetEntryAssembly` is not always correct) and other already eval'd code (but only those used already within current request/context), also after the compilation it should denotate what on what eval's it depends on... this needs some additions in the compiler itself.

jakubmisek:
this should be fixed with this line of code

https://github.com/peachpiecompiler/peachpie/blob/f46152df76c034d9bbf30a3c8231dbe062cadec5/src/Peachpie.Library.Scripting/PhpCompilationFactory.cs#L75

calvinbaart:
Properly fixed by @jakubmisek

Issue #396, opened 2019-03-29
System.InvalidOperationException: Call to undefined function: 'password_hash'


menturion:
Got this exception `System.InvalidOperationException: Call to undefined function: 'password_hash'`

-> Function "password_hash" (s. https://www.php.net/manual/en/function.password-hash.php) is not supported.

See related issues; https://github.com/peachpiecompiler/peachpie/issues/395 and https://github.com/peachpiecompiler/peachpie/issues/379

Issue #395, opened 2019-03-29
System.InvalidOperationException: 'Call to undefined function: 'password_needs_rehash'


menturion:
Due to PeachPie still does not support "password_verify" (see issue https://github.com/peachpiecompiler/peachpie/issues/379) I bypassed the password check for testing purposes which results into the following exception: 

`System.InvalidOperationException: 'Call to undefined function: 'password_needs_rehash'`

-> Function "password_needs_rehash" (s. https://www.php.net/manual/en/function.password-needs-rehash.php) is also not supported.

See related issues; https://github.com/peachpiecompiler/peachpie/issues/396

PR #501, opened 2019-08-24
Dev/password hash

TomatorCZ:
Fixes https://github.com/peachpiecompiler/peachpie/issues/395
Fixes https://github.com/peachpiecompiler/peachpie/issues/396

@roberthusak  
**Changes:**
_trevis.yml_: password_hash uses Argon2 algorithm. This Algorithm is supported in php version 7.2+
_unset_002.php_: This test using (unset) cast, which is already deprecated in new version of PHP ([https://www.php.net/manual/en/language.types.null.php#language.types.null.casting])
_bug_33853.php_: change __autoload function for spl_autoloadregister, because is already deprecated ([https://www.php.net/manual/en/function.autoload.php])
other changed tests : Suppress error DEPRECATED 
**Notice:**
- There is verifying  argon2 hash in password_verify function instead of crypt function because in php crypt does not support this algorithm


jakubmisek:
nic! can we have `password_needs_rehash` implemented as well?

TomatorCZ:
Ok

src/Peachpie.Library/Hash.cs
```diff
@@ -2618,16 +2622,265 @@ public static PhpString hash_pbkdf2(string algo, byte[] password, byte[] salt, i
...
+
+        private static PhpValue HashPasswordBlowfish(string password,PhpArray opt)
+        {          
+            // Default setting for bcrypt
+            int cost = costDefault;
+            string salt = BCrypt.Net.BCrypt.GenerateSalt(cost);
+
+            if (opt != null)
+            {
+                PhpValue pom;
```
roberthusak: I suggest removing the variable ```pom``` from here and creating separate variables with appropriate names (e.g. ```costValue``` and ```saltValue```) on lines 2736 and 2759.

src/Peachpie.Library/Hash.cs
```diff
@@ -2618,16 +2622,265 @@ public static PhpString hash_pbkdf2(string algo, byte[] password, byte[] salt, i
...
+
+                    }
+                    else
+                        result = true;
+                    break;
+                // Argon2i
+                case 2:
+                // Argon2id
+                case 3:
+                    Regex reg = new Regex(@"^\$(argon2id|argon2id)\$v=\d+\$m=(\d+),t=(\d+),p=(\d+)\$.+\$.+$");
```
roberthusak: Create a static field from ```reg``` so that it doesn't have to be recreated with each call to the function.

tests/hash/password_hash_001.php
```diff
@@ -0,0 +1,16 @@
...
+    'cost' => 15,
+    'salt' => $salt,
+];
+
+$hashD = @password_hash($password,PASSWORD_DEFAULT,$options);
+$hashB = @password_hash($password,PASSWORD_BCRYPT,$options);
+
+echo "PASSWORD_DEFAULT : $hashD\n";
+echo "PASSWORD_BCRYPT : $hashB\n";
+?>
```
roberthusak: We usually don't put the ```?>``` closing tags in tests, remove it from all the added ones, please. In ```.phpt``` files in the PHP source code they do it to separate the code execution from the following lines with the expected results, we don't need to do that.

src/Peachpie.Library/Hash.cs
```diff
@@ -2618,16 +2622,265 @@ public static PhpString hash_pbkdf2(string algo, byte[] password, byte[] salt, i
...
+                            cost = pomInt;
+                        else
+                        {
+                            PhpException.Throw(PhpError.Warning, Resources.LibResources.bcrypt_invalid_cost, pomInt.ToString());
+                            return PhpValue.False;
+                        }
+                    }
+                    else
+                    {
+                        PhpException.Throw(PhpError.Warning, Resources.LibResources.bcrypt_undefined_const, pom.ToString()); // invalid value of cost
```
jakubmisek: we already have a resource string for undefined constants
```
Core.Resources.ErrResources.constant_not_found
```

src/Peachpie.Library/Hash.cs
```diff
@@ -2618,16 +2622,264 @@ public static PhpString hash_pbkdf2(string algo, byte[] password, byte[] salt, i
...
+        private static PhpValue HashPasswordBlowfish(string password,PhpArray opt)
+        {          
+            // Default setting for bcrypt
+            int cost = costDefault;
+            string salt = BCrypt.Net.BCrypt.GenerateSalt(cost);
+
+            if (opt != null)
+            {
+                // Argument cost for bcrypt
+                if (opt.ContainsKey("cost"))
```
jakubmisek: please replace with pattern
```c#
if (opt.TryGetValue("cost", out var costValue)) { ...
```

src/Peachpie.Library/Hash.cs
```diff
@@ -2618,16 +2622,264 @@ public static PhpString hash_pbkdf2(string algo, byte[] password, byte[] salt, i
...
+                {
+                    PhpValue costValue = opt.GetItemValue(new IntStringKey("cost"));
+                    if (costValue.IsInteger()) // Check right value of argument
```
jakubmisek: please replace with pattern
```c#
if (costValue.IsLong(out var costInt)) { ...
```
also consider what if `costValue` is `double` ? I suppose it would work in PHP, but we would throw a warning instead.

src/Peachpie.Library/Hash.cs
```diff
@@ -2618,16 +2622,264 @@ public static PhpString hash_pbkdf2(string algo, byte[] password, byte[] salt, i
...
+                    }
+                    else
+                    {
+                        PhpException.Throw(PhpError.Warning, Core.Resources.ErrResources.constant_not_found, costValue.ToString()); // invalid value of cost
+                        return PhpValue.False;
+                    }
+                }
+                
+                // Argument salt for bcrypt
+                if (opt.ContainsKey("salt"))
```
jakubmisek: `TryGetValue` as above

src/Peachpie.Library/Hash.cs
```diff
@@ -2618,16 +2622,264 @@ public static PhpString hash_pbkdf2(string algo, byte[] password, byte[] salt, i
...
+                // Default
+                case 0:
+                // BCrypt
+                case 1:
+                    string[] hashParts = hash.Split('$');
+
+                    if (hashParts.Length >= 3 && hashParts[1][0] == '2' && hashParts[1].Length >= 2) // $2 $ Right algorithm
+                    {
+
+                        if (opt != null && opt.ContainsKey("cost")) // Check options
```
jakubmisek: replace with `TryGetValue("cost", out var pom)`

src/Peachpie.Library/Hash.cs
```diff
@@ -2618,16 +2622,264 @@ public static PhpString hash_pbkdf2(string algo, byte[] password, byte[] salt, i
...
+                // Argon2id
+                case 3:
+                    Match match = expressionHashArgon2.Match(hash);
+                    
+                    if (match != null && (algo == 2 && match.Groups[1].Value == "argon2i") || (algo == 3 && match.Groups[1].Value == "argon2id")) // Check right algorithm
+                    {
+                        if (opt != null) // Check options
+                        {
+
+                            if (opt.ContainsKey("memory_cost"))
```
jakubmisek: `TryGetValue` as noted above

src/Peachpie.Library/Hash.cs
```diff
@@ -2618,16 +2622,264 @@ public static PhpString hash_pbkdf2(string algo, byte[] password, byte[] salt, i
...
+                                result = !(opt.GetItemValue(new IntStringKey("memory_cost")) == match.Groups[2].Value);
+
+                            if (opt.ContainsKey("time_cost"))
```
jakubmisek: `TryGetValue` as noted above

src/Peachpie.Library/Hash.cs
```diff
@@ -2618,16 +2622,264 @@ public static PhpString hash_pbkdf2(string algo, byte[] password, byte[] salt, i
...
+                                result = result || !(opt.GetItemValue(new IntStringKey("time_cost")) == match.Groups[3].Value);
+
+                            if (opt.ContainsKey("time_cost"))
```
jakubmisek: `TryGetValue` as noted above

src/Peachpie.Library/Hash.cs
```diff
@@ -2618,16 +2622,264 @@ public static PhpString hash_pbkdf2(string algo, byte[] password, byte[] salt, i
...
+        }
 
         /// <summary>
         /// Verifies that a password matches a hash.
         /// </summary>
         public static bool password_verify(string password, string hash)
         {
+            try
+            {
+                if (hash.Substring(1, 7) == "argon2i")
```
jakubmisek: check hash string lengths first! avoid throwing OutOfRange exception

jakubmisek: anyways the point is to avoid throwing `OutOfRangeException`

src/Peachpie.Library/Hash.cs
```diff
@@ -2618,14 +2622,251 @@ public static PhpString hash_pbkdf2(string algo, byte[] password, byte[] salt, i
...
+            // Default setting for bcrypt
+            long cost = costDefault;
+            string salt = BCrypt.Net.BCrypt.GenerateSalt((int)cost);
+
+            if (opt != null)
+            {
+                // Argument cost for bcrypt
+                if (opt.TryGetValue("cost", out var costValue))
+                {
+                    if (costValue.IsLong(out var costInt)) // Check right value of argument
```
jakubmisek: what if `costValue` is `double` ? Does it work in PHP?

src/Peachpie.Library/Hash.cs
```diff
@@ -2618,14 +2622,251 @@ public static PhpString hash_pbkdf2(string algo, byte[] password, byte[] salt, i
...
+                            cost = costInt;
+                        else
+                        {
+                            PhpException.Throw(PhpError.Warning, Resources.LibResources.bcrypt_invalid_cost, costInt.ToString());
+                            return PhpValue.False;
+                        }
+                    }
+                    else
+                    {
+                        PhpException.Throw(PhpError.Warning, Core.Resources.ErrResources.constant_not_found, costValue.ToString()); // invalid value of cost
```
jakubmisek: Does PHP report "constant_not_found" if `cost` is `double` ?

src/Peachpie.Library/Hash.cs
TomatorCZ: Yes, value is trancuated, it can be also string. Can I use method ToInt() on PhpValue ?

PR #563, opened 2019-10-22
Speed up building and testing in Travis CI

roberthusak:
- Add namespaces to all the tests in `tests` folder so that they are isolated.
- Fix the tests accordingly (mostly by prepending library types by `\` and using the particular namespace in reflection stuff).
- Fix the usage of `__FUNCTION__` inside a namespace.
- Build the tests as one project in Travis CI and only run them one by one - about 4x times faster than compiling them separately.
- Run Peachpie.Runtime.Tests in Travis CI.


tests/arrays/array_map.php
```diff
@@ -18,8 +19,8 @@ function g($x)
   $d = array('A' => 1,10 => 2);
   
   print_r(array_map(null,$a,$b,$c));
-  print_r(array_map("f",$a,$b,$c));
-  print_r(array_map("g",$d));
+  print_r(array_map("arrays\\array_map\\f",$a,$b,$c));
```
jakubmisek: a pseudoconstant would be nicer `__NAMESPACE__ . "\\f"`

tests/program.php
```diff
@@ -0,0 +1,3 @@
+<?php
+
+include $_SERVER['argv'][1];
```
jakubmisek: isn't it just `$argv[1]` ?

tests/program.php
```diff
@@ -0,0 +1,3 @@
+<?php
```
jakubmisek: if you'd rename it to `program.phpx` or something, you can get rid of most of the workarounds ;)

tests/program.php
roberthusak: It seemed not to work before, but now it does. Maybe it was due to some other reason that I missed.. Anyway, I changed it.

jakubmisek:
thank you @roberthusak !

tests/program.php
roberthusak: Seems a bit hacky, but useful indeed :grin:

roberthusak:
@jakubmisek ok, thanks for the review!

PR #567, opened 2019-10-24
Update ImageSharp to latest beta

JimBobSquarePants:
- Updates all Sixlabors dependencies to the latest betas.
- Refactors graphics API calls to reflect changes.
- Adds temporary base `ImageProcessor<TPixel>` class to correctly handle per-frame manipulation.
- Changes loading behaviour to only preserve first frame of multi-frame image.
- Optimizes general EXIF metadata retrieval.

Note: I couldn't find unit tests for the graphics functionality so my changes are untested.

### DIFF
--- a/src/Peachpie.Library.Graphics/PhpGd2.cs
+++ b/src/Peachpie.Library.Graphics/PhpGd2.cs
@@ -426,7 +426,7 @@ namespace Peachpie.Library.Graphics
 
             try
             {
-                return new PhpGdImageResource(Image.Load(image, out var format), format);
+                return new PhpGdImageResource(Image.Load<Rgba32>(image, out var format), format);
             }
             catch
             {
@@ -535,7 +535,7 @@ namespace Peachpie.Library.Graphics
             {
                 if (stream != null)
                 {
-                    try { img = Image.Load(configuration, stream, out format); }
+                    try { img = Image.Load<Rgba32>(configuration, stream, out format); }
                     catch { }
                 }
             }
@@ -1009,7 +1009,7 @@ namespace Peachpie.Library.Graphics
                 return false;
             }
 
-            img.tiled = new ImageBrush<Rgba32>(imgTile.Image);
+            img.tiled = new ImageBrush(imgTile.Image);
 
             return false;
         }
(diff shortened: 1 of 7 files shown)


src/Peachpie.Library.Graphics/PhpGd2.cs
```diff
@@ -426,7 +426,18 @@ public static PhpResource imagecreatefromstring(byte[] image)
...
+                var img = Image.Load<Rgba32>(image, out var format);
+
+                // PHP will only preserve the first frame of a multi-frame image.
+                if (img.Frames.Count > 0)
+                {
+                    for (int i = 1; i < img.Frames.Count; i++)
+                    {
+                        img.Frames.RemoveFrame(i);
+                    }
+                }
```
jakubmisek: shuldn't it be more like:
```
while (img.Frames.Count > 1)
{
  img.Frames.RemoveFrame(img.Frames.Count - 1);
}
```
otherwise it would remove just the every second frame, and every removed frame would cause an unnecessary copying (since removing from 1 to count)

src/Peachpie.Library.Graphics/PhpGd2.cs
```diff
@@ -535,7 +546,20 @@ static PhpGdImageResource imagercreatefromfile(Context ctx, string filename, ICo
...
+                        img = Image.Load<Rgba32>(configuration, stream, out format);
+
+                        // PHP will only preserve the first frame of a multi-frame image.
+                        if (img.Frames.Count > 0)
+                        {
+                            for (int i = 1; i < img.Frames.Count; i++)
+                            {
+                                img.Frames.RemoveFrame(i);
+                            }
+                        }
```
jakubmisek: code duplicity, will be moved into a helper

src/Peachpie.Library.Graphics/PhpGd2.cs
JimBobSquarePants: Whoops! Yes, should be `>1’ that should teach me to write code after midnight!

jakubmisek: Haha and there shouldn't be `i++` :)) ANYWAYS thanks for the pull request!

PR #571, opened 2019-10-28
Dev/ftp

TomatorCZ:
Please @roberthusak  for review
Completed all functions in ftp module. I added non blocking functions and few common functions from last PR. 
Issue with startpos and resumepos parameter in ftp_get,put,fput,fget etc. is not solved. I wrote issue to report this feature [https://github.com/robinrodricks/FluentFTP/issues/456](url)


src/Peachpie.Library/Ftp.cs
```diff
@@ -875,5 +881,421 @@ public static bool ftp_site(PhpResource ftp_stream, string command)
...
+                if (resumepos == FTP_AUTORESUME)
+                {
+                    stream.Seek(0, SeekOrigin.End);
+                    resumepos = stream.Tell();
+                }
+                else
+                {
+                    stream.Seek(resumepos, SeekOrigin.Begin);
+                }
+            }
```
roberthusak: It's the same logic as in `ftp_fget`, wouldn't it be better to move it to a common helper method? Not only for code size, but also for maintainability (e.g. if we need to change the logic, we can do it in one place).

src/Peachpie.Library/Resources/Resources.resx
```diff
@@ -681,4 +681,7 @@
   <data name="pcre_make_error" xml:space="preserve">
     <value>{0} at offset {1}</value>
   </data>
+  <data name="ftp_error_no_nb" xml:space="preserve">
+    <value>No nbronous transfer to continue.</value>
```
roberthusak: Put _asynchronous_ there - unlike PHP we are trully asynchronous and _nbronous_ is a really weird word :-)

src/Peachpie.Library/Ftp.cs
```diff
@@ -65,6 +72,40 @@ public FtpResource(FtpClient client, int timeout)
 
             public bool Autoseek { get; set; }
 
+            public Task<bool> TaskFunction { get; set; }
```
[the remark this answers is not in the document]
jakubmisek: What does it even mean, `Function` ? Is it a `PendingTask` or `PendingAsyncOperationTask` ?

src/Peachpie.Library/Ftp.cs
```diff
@@ -675,6 +682,17 @@ public static long ftp_mdtm(PhpResource ftp_stream, string remote_file)
...
         }
 
+        /// <summary>
+        /// Converts octal number, which represents mode, from input to same value in decimal base.
+        /// </summary>
+        /// <param name="octalMode"></param>
+        /// <returns>Same value in decimal base</returns>
+        private static int ConvertUnixModeFromInput(int octalMode)
+        {
+            // TODO: Write specificum of FLuent ftp...
```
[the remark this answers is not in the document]
jakubmisek: yep write: "FluentFtp converts given number to string and sends it as it is - this means the mode digits must be encoded as decimal numbers, not octal as it comes from PHP"

src/Peachpie.Library/Resources/Resources.resx
roberthusak: @jakubmisek If it does, I wasn't able to find it. The docs of the `ftp_nb_*` functions always mention only _asynchronous_ or _non-blocking_, so this word doesn't make any sense. If we want to keep closer to the original, I'd suggest: "No asynchronous (non-blocking) transfer to continue."

PR #610, opened 2020-01-13
Fix handling of invalid callbacks and enhance type analysis of callable parameters

### LINKED ISSUE #607: Invoking invalid callback doesn't throw exception
Test case:
```php
function test($callback) {
  try {
    $callback();
    echo "OK";
  }
  catch (Throwable $e) {
    echo get_class($e);
  }
}

test("foo");
```
Expected: `Error`
Actual: `OK`

### LINKED ISSUE #608: Passing invalid callback as callable argument doesn't throw exception
Test case:
```php
function check_callable(callable $callback) {}

function test($callback) {
  try {
    check_callable($callback);
    echo "OK";
  }
  catch (Throwable $e) {
    echo get_class($e);
  }
}

test("foo");
```
Expected: `TypeError`
Actual: `OK`

### LINKED ISSUE #609: Use callable parameter type hint for type analysis
The following test put in `Peachpie.DiagnosticTests` should pass:
```php
function callable_check(callable /*|array|Closure|object|string|*/$c) {}
```
Actual result: `Wrong type mixed instead of array|Closure|object|string`

roberthusak:
Fixes #607
Fixes #608
Fixes #609

### DIFF
--- a/src/Peachpie.CodeAnalysis/CodeGen/VariableReference.cs
+++ b/src/Peachpie.CodeAnalysis/CodeGen/VariableReference.cs
@@ -839,7 +839,7 @@ namespace Pchp.CodeAnalysis.Semantics
 
         static void EmitTypeCheck(CodeGenerator cg, IPlace valueplace, SourceParameterSymbol srcparam)
         {
-            // TODO: check callable, iterable, type if not resolved in ct
+            // TODO: check iterable, type if not resolved in ct
 
             // check NotNull
             if (srcparam.IsNotNull)
@@ -864,6 +864,19 @@ namespace Pchp.CodeAnalysis.Semantics
                     cg.Builder.EmitIntConstant(srcparam.ParameterIndex + 1);
                     cg.EmitPop(cg.EmitCall(ILOpCode.Call, cg.CoreMethods.Operators.ThrowIfArgumentNull_object_int));
                 }
+            }
+
+            // check callable
+            if (srcparam.Syntax.TypeHint.IsCallable())
+            {
+                cg.EmitSequencePoint(srcparam.Syntax);
+
+                // Template: PhpException.ThrowIfArgumentNotCallable(<ctx>, current RuntimeTypeHandle, value, arg)
+                cg.EmitLoadContext();
+                cg.EmitCallerTypeHandle();
+                cg.EmitConvertToPhpValue(valueplace.EmitLoad(cg.Builder), default);     // To handle conversion from PhpAlias when the parameter is by ref
+                cg.Builder.EmitIntConstant(srcparam.ParameterIndex + 1);
+                cg.EmitPop(cg.EmitCall(ILOpCode.Call, cg.CoreMethods.Operators.ThrowIfArgumentNotCallable_Context_RuntimeTypeHandle_PhpValue_int));
             }
         }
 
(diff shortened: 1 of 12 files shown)


src/Peachpie.CodeAnalysis/CodeGen/VariableReference.cs
```diff
@@ -865,6 +865,20 @@ static void EmitTypeCheck(CodeGenerator cg, IPlace valueplace, SourceParameterSy
                     cg.EmitPop(cg.EmitCall(ILOpCode.Call, cg.CoreMethods.Operators.ThrowIfArgumentNull_object_int));
                 }
             }
+
+            // check callable
+            if (srcparam.Syntax.TypeHint.IsCallable(out bool isNullable))
```
jakubmisek: `isNullable` is not correct, do not use it. Parameter can also be `null` if its default value is `NULL`. Use `srcparam.IsNotNull` instead.
```php
function foo(callable $c = null) { } // <-- $c can be NULL
```

src/Peachpie.CodeAnalysis/Symbols/Php/PhpRoutineSymbolExtensions.cs
```diff
@@ -144,6 +144,15 @@ public static TypeRefMask GetResultType(this IPhpValue symbol, TypeRefContext ct
                     Debug.Assert(t.IsSZArray());
                     return ctx.GetArrayTypeMask(TypeRefFactory.CreateMask(ctx, ((ArrayTypeSymbol)t).ElementType));
                 }
+                else if (ps.Syntax.TypeHint.IsCallable(out bool isNullable))
```
jakubmisek: `isNullable` is not correct, do not use it. Parameter can also be NULL is its default is `NULL`, use `ps.IsNotNull` instead

src/Peachpie.CodeAnalysis/Utilities/AstUtils.cs
```diff
@@ -304,6 +304,25 @@ public static bool IsNullable(this TypeRef tref)
             return tref is NullableTypeRef; // && tref != null
         }
 
+        /// <summary>
+        /// Gets value indicating the type refers to <c>callable</c> or <c>?callable</c>.
+        /// </summary>
+        public static bool IsCallable(this TypeRef tref, out bool isNullable)
```
jakubmisek: `out bool isNullable` can be removed, causes issues

src/Peachpie.CodeAnalysis/CodeGen/VariableReference.cs
```diff
@@ -865,6 +865,20 @@ static void EmitTypeCheck(CodeGenerator cg, IPlace valueplace, SourceParameterSy
                     cg.EmitPop(cg.EmitCall(ILOpCode.Call, cg.CoreMethods.Operators.ThrowIfArgumentNull_object_int));
                 }
             }
+
+            // check callable
+            if (srcparam.Syntax.TypeHint.IsCallable())
+            {
+                cg.EmitSequencePoint(srcparam.Syntax);
+
+                // Template: PhpException.ThrowIfArgumentNotCallable(<ctx>, current RuntimeTypeHandle, value, arg)
```
jakubmisek: seems not correct, missing `bool` argument

PR #611, opened 2020-01-15
Add several semantic transformations

roberthusak:
The main contribution are the following semantic transformations:

- Removal of unnecessary copying:
  - Passing values of routine parameters
  - Copying local variables assigned between each other
  - Copying return values
- ```ord($s[$i])``` to ```(int)s[i]``` (not creating an intermediate one-byte string)
- ```isset($a[$i]) ? $a[$i] : fallback``` to ```TryGetItemValue($a, $i, out val) ? val : fallback```
- ```-1 * X```, ```X * -1``` to ```-X```
- ```empty($x)``` to false if ```$x``` is not initialized
- ```define("X", VALUE)``` to ```const X = VALUE```
- ```(IPhpCallable)[$this, "method"]```, ```(IPhpCallable)["SomeClass", "method"]```, ```(IPhpCallable)"someFunction"``` is now resolved in compiled time and cached in runtime

Related fixes and enhancements:

- The analysis reducing numbers to ```int``` works now for reverse loops as well (e.g. ```for ($i = 10; $i > 0; $i--) { ... }```).
- Added ```FixpointAnalysis<TContext, TState>``` for lightweight intraprocedural analyses (separated from the global analysis performed in the analysis phase), currently used to remove unnecessary copying.
- Introduced ```SingleBlockWalker<T>``` as a step between ```GraphVisitor<T>``` and ```GraphWalker<T>```.
- Updated ```BuilderVisitor``` to produce CFG with continuous block ordinals (some of the "allocated" ordinals weren't originally used by being overwritten).
- Other smaler updates.


src/Peachpie.CodeAnalysis/CodeGen/Graph/BoundExpression.cs
```diff
@@ -2345,7 +2345,12 @@ internal override TypeSymbol Emit(CodeGenerator cg)
                 else
                 {
                     Debug.Assert(Receiver != null);
-                    throw cg.NotImplementedException();
+
+                    // methodRoutineInfo.BindTarget(Receiver)
+                    m.EmitLoadRoutineInfo(cg);
+                    cg.EmitConvert(Receiver, cg.CoreTypes.Object);
```
jakubmisek: EmitConvert(`Receiver`,  `T`)

src/Peachpie.CodeAnalysis/CodeGen/Graph/BoundExpression.cs
```diff
@@ -4973,6 +5022,85 @@ internal override TypeSymbol Emit(CodeGenerator cg)
...
+                {
+                    Debug.Assert(indexType == cg.CoreTypes.PhpValue);
+                    operation = cg.CoreMethods.Operators.TryGetItemValue_PhpArray_PhpValue_PhpValueRef;
+                }
+            }
+            else
+            {
+                Debug.Assert(arrType == cg.CoreTypes.PhpValue);
+                Debug.Assert(indexType == cg.CoreTypes.PhpValue);
+                operation = cg.CoreMethods.Operators.TryGetItemValue_PhpValue_PhpValue_PhpValueRef;
```
jakubmisek: question: does it handle `ArrayAccess` instances with `offsetGet` and `offsetExists` methods ?

src/Peachpie.CodeAnalysis/CodeGen/Symbols/MethodSymbol.cs
```diff
@@ -6,29 +6,58 @@
...
+                {
+                    overloads = ImmutableArray.Create(method);
+                }
+
+                // CreateUserRoutine("method", new[] { (MethodInfo)MethodBase.GetMethodFromHandle(ldtoken method), ... })
+                il.EmitStringConstant(method.MetadataName);
+                cg.Emit_NewArray(cg.CoreTypes.MethodInfo, overloads, m =>
+                {
+                    cg.EmitLoadToken(m, null);
+                    cg.EmitCall(ILOpCode.Call, cg.CoreMethods.Reflection.GetMethodFromHandle_RuntimeMethodHandle);
```
jakubmisek: GetWellKnownMethod( ... ) instead

src/Peachpie.CodeAnalysis/CodeGen/Symbols/MethodSymbol.cs
```diff
@@ -6,29 +6,58 @@
...
+                    cg.EmitCastClass(cg.CoreTypes.MethodInfo);
```
jakubmisek: GetWellKnownType( ... ) instead

src/Peachpie.CodeAnalysis/CodeGen/VariableReference.cs
```diff
@@ -1110,6 +1110,11 @@ public void Store(CodeGenerator cg, TypeSymbol valuetype)
 
         #endregion
 
+        /// <summary>
+        /// Whether to skip passing the parameter value, as it is unnecessary (e.g. unused parameter or only delegation to another method)
+        /// </summary>
+        public bool SkipPass { get; set; }
```
jakubmisek: it shouldn't be in `VariableReference` ... 

can we just refer to the FlowState (as we do with types)?

src/Peachpie.CodeAnalysis/CodeGen/VariableReference.cs
```diff
@@ -1176,8 +1181,11 @@ public override void EmitInit(CodeGenerator cg)
 
             if (source == target)
             {
-                // 2a. (source == target): Pass (inplace copy)
-                source.EmitPass(cg);
+                // 2a. (source == target): Pass (inplace copy and dereference) if necessary
+                if (!SkipPass)
```
jakubmisek: do not check it in here, add the condition inside `EmitPass` ...

src/Peachpie.CodeAnalysis/FlowAnalysis/ExpressionAnalysis.cs
```diff
@@ -164,12 +164,26 @@ public override T VisitStaticStatement(BoundStaticVariableStatement x)
                 // analyse initializer
                 Accept(v.InitialValue);
 
-                State.SetLessThanLongMax(local, (v.InitialValue.ConstantValue.HasValue && v.InitialValue.ConstantValue.Value is long && (long)v.InitialValue.ConstantValue.Value < long.MaxValue));
+                if (v.InitialValue.ConstantValue.IsInteger(out long val))
+                {
+                    if (val < long.MaxValue)
```
jakubmisek: else ?

src/Peachpie.CodeAnalysis/FlowAnalysis/ExpressionAnalysis.cs
```diff
@@ -164,12 +164,26 @@ public override T VisitStaticStatement(BoundStaticVariableStatement x)
...
+                        State.SetLessThanLongMax(local, true);
+
+                    if (val > long.MinValue)
```
jakubmisek: else ?

src/Peachpie.CodeAnalysis/FlowAnalysis/ExpressionAnalysis.cs
```diff
@@ -164,12 +164,26 @@ public override T VisitStaticStatement(BoundStaticVariableStatement x)
                 // analyse initializer
                 Accept(v.InitialValue);
 
-                State.SetLessThanLongMax(local, (v.InitialValue.ConstantValue.HasValue && v.InitialValue.ConstantValue.Value is long && (long)v.InitialValue.ConstantValue.Value < long.MaxValue));
```
jakubmisek: just keep this line, try to not nest everything in several `if`s

src/Peachpie.CodeAnalysis/Semantics/Graph/BuilderVisitor.cs
```diff
@@ -259,18 +259,20 @@ private void AddFinalReturn()
             return block;
         }
 
-        private CatchBlock/*!*/NewBlock(CatchItem item)
+        private BoundBlock/*!*/NewUnorderedBlock() => new BoundBlock();
```
jakubmisek: just rename it back please

src/Peachpie.CodeAnalysis/Utilities/BitMask.cs
```diff
@@ -0,0 +1,79 @@
...
+namespace Peachpie.CodeAnalysis.Utilities
+{
+    /// <summary>
+    /// A general structure to record bit information about indexed entities in flow analysis.
+    /// </summary>
+    /// <remarks>
+    /// The minimum and initial lattice element is 0, the maximum is ~0. Indices outside the scope of ulong (64)
+    /// are regarded as present in the set.
+    /// </remarks>
+    internal struct BitMask
```
jakubmisek: isn't there enough BitMasks already?

src/Peachpie.CodeAnalysis/Utilities/FixpointAnalysis.cs
```diff
@@ -0,0 +1,104 @@
...
+using Pchp.CodeAnalysis.Symbols;
+using Pchp.CodeAnalysis.Utilities;
+
+namespace Peachpie.CodeAnalysis.Utilities
+{
+    /// <summary>
+    /// Interface to define custom intraprocedural fixpoint analyses.
+    /// </summary>
+    /// <typeparam name="TState">Flow state to propagate throughout the graph.</typeparam>
+    internal interface IFixPointAnalysisContext<TState>
```
jakubmisek: why is it in `Utilities` ?

src/Peachpie.Runtime/Operators.cs
```diff
@@ -614,6 +614,66 @@ public static string GetItemValue(string value, PhpValue index, bool quiet)
...
+        {
+            switch (value.TypeCode)
+            {
+                case PhpTypeCode.String:
+                    return GetItemOrdValue(value.String, index);
+                case PhpTypeCode.MutableString:
+                    return GetItemOrdValue(value.MutableString, index);
+                default:
+                    var item = value.GetArrayItem(index);
+                    var itemType = item.TypeCode;
```
jakubmisek: use `IsPhpString()` instead

src/Peachpie.Runtime/Operators.cs
```diff
@@ -657,6 +717,55 @@ public static PhpAlias EnsureItemAlias(this IPhpArray array, PhpValue index, boo
...
+            else
+            {
+                item = default;
+                return false;
+            }
+        }
+
+        public static bool TryGetItemValue(this PhpValue value, PhpValue index, out PhpValue item)
+        {
+            if (value.Object is PhpArray array)
```
jakubmisek: `value.Object is PhpAlias alias` ?

src/Peachpie.Runtime/Reflection/PhpRoutineInfo.cs
```diff
@@ -63,6 +63,10 @@ public abstract class RoutineInfo : IPhpCallable, ICloneable
         /// <summary>Target instance when binding the MethodInfo call.</summary>
         internal virtual object Target => null;
 
+        /// <summary>Creates an <see cref="IPhpCallable"/> from an instance method, binding the target to call the method on.</summary>
+        /// <remarks>Returns <c>null</c> if not applicable.</remarks>
+        public virtual IPhpCallable BindTarget(object target) => null;
```
jakubmisek: remove

whenever needed, use 
```c#
PhpCallback.Create(target, callable)
```

src/Peachpie.Runtime/Reflection/PhpRoutineInfo.cs
```diff
@@ -290,6 +294,30 @@ public PhpMethodInfoWithBoundType(int index, string name, MethodInfo[] methods,
 
         #endregion
 
+        #region BoundTargetCallable
```
jakubmisek: remove, move to `PhpCallback`

src/Peachpie.Runtime/Reflection/PhpRoutineInfo.cs
```diff
@@ -335,6 +363,8 @@ static bool AllDeclaredInBase(MethodInfo[] methods, Type callertype)
 
         public override PhpTypeInfo DeclaringType => _methods[0].DeclaringType.GetPhpTypeInfo();
 
+        public override IPhpCallable BindTarget(object target) => new BoundTargetCallable(target, PhpInvokable);
```
jakubmisek: remove

src/Peachpie.CodeAnalysis/Semantics/Graph/SingleBlockWalker.cs
```diff
@@ -0,0 +1,561 @@
...
+using System.Diagnostics;
+using System.Text;
+using Pchp.CodeAnalysis.Semantics.TypeRef;
+
+namespace Pchp.CodeAnalysis.Semantics.Graph
+{
+    /// <summary>
+    /// Visitor used to traverse a single block and its adjacent edge.
+    /// </summary>
+    public abstract class SingleBlockWalker<T> : GraphVisitor<T>
```
jakubmisek: another visitor?

src/Tests/Peachpie.DiagnosticTests/tests/type_analysis_09.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+class C {}
+
+function foo(array $a, C $c, ?C $cn1, C $cn2 = null) {}
+
+function bar($a, $c, $cn1, $cn2) {
+  foo(/*|mixed|*/$a, /*|mixed|*/$c, /*|mixed|*/$cn1, /*|mixed|*/$cn2);
+  foo(/*|array|*/$a, /*|C|*/$c, /*|C|null|*/$cn1, /*|C|null|*/$cn2);
```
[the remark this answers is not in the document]
roberthusak: No, according to the [manual](https://www.php.net/manual/en/language.types.type-juggling.php):

> The behaviour of an automatic conversion to array is currently undefined.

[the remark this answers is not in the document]
roberthusak: I also tested it manually and all the other types caused ```TypeError```.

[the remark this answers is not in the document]
jakubmisek: can we ensure, we work properly? If `$a` become annotated as `array` while it isn't, we'll be emitting incorrect IL then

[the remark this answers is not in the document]
roberthusak: I see.. Added a test in 095e86162da2163bf2ab4cf61b5aaaf7696c2617, is it enough?

src/Peachpie.CodeAnalysis/Symbols/CoreType.cs
```diff
@@ -169,6 +169,8 @@ public CoreTypes(PhpCompilation compilation)
             Exception = CreateFromFullName(WellKnownTypes.GetMetadataName(WellKnownType.System_Exception));
             RuntimeTypeHandle = Create(SpecialType.System_RuntimeTypeHandle);
             RuntimeMethodHandle = Create(SpecialType.System_RuntimeMethodHandle);
+            MethodBase = CreateFromFullName("System.Reflection.MethodBase");
+            MethodInfo = CreateFromFullName("System.Reflection.MethodInfo");
```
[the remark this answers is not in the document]
roberthusak: Updated in 1c87acd2ce3b892f3abef5337cef31cf3451e891.

src/Peachpie.CodeAnalysis/Symbols/CoreMembers.cs
```diff
@@ -960,6 +968,10 @@ public ReflectionHolder(CoreTypes ct)
             {
                 _ct = ct;
                 _lazyCreateUserRoutine = null;
+                _lazyCreateUserRoutineMethod = null;
+
+                BindTarget_Object = ct.RoutineInfo.Method("BindTarget", ct.Object);
+                GetMethodFromHandle_RuntimeMethodHandle = ct.MethodBase.Method("GetMethodFromHandle", ct.RuntimeMethodHandle);
```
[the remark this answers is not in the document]
roberthusak: Updated in 1c87acd2ce3b892f3abef5337cef31cf3451e891.

src/Peachpie.CodeAnalysis/Symbols/CoreType.cs
```diff
@@ -169,6 +169,8 @@ public CoreTypes(PhpCompilation compilation)
             Exception = CreateFromFullName(WellKnownTypes.GetMetadataName(WellKnownType.System_Exception));
             RuntimeTypeHandle = Create(SpecialType.System_RuntimeTypeHandle);
             RuntimeMethodHandle = Create(SpecialType.System_RuntimeMethodHandle);
+            MethodBase = CreateFromFullName("System.Reflection.MethodBase");
```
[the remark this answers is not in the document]
roberthusak: Updated in 1c87acd2ce3b892f3abef5337cef31cf3451e891.

src/Peachpie.CodeAnalysis/Symbols/CoreMembers.cs
```diff
@@ -960,6 +968,10 @@ public ReflectionHolder(CoreTypes ct)
             {
                 _ct = ct;
                 _lazyCreateUserRoutine = null;
+                _lazyCreateUserRoutineMethod = null;
```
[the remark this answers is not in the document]
roberthusak: Updated in 1c87acd2ce3b892f3abef5337cef31cf3451e891.

src/Peachpie.CodeAnalysis/CodeGen/Symbols/MethodSymbol.cs
roberthusak: Updated in 1c87acd2ce3b892f3abef5337cef31cf3451e891.

src/Peachpie.CodeAnalysis/CodeGen/Symbols/MethodSymbol.cs
roberthusak: Updated in 1c87acd2ce3b892f3abef5337cef31cf3451e891.

src/Peachpie.CodeAnalysis/FlowAnalysis/Passes/TransformationRewriter.CopyAnalysis.cs
```diff
@@ -0,0 +1,352 @@
...
+    partial class TransformationRewriter
+    {
+        /// <summary>
+        /// Each state in copy analysis maps each variable to a set of <see cref="BoundCopyValue"/> instances.
+        /// Whenever a variable is modified, all the mapped copy operations are considered unavailable for removal
+        /// due to the possible aliasing.
+        /// </summary>
+        private struct CopyAnalysisState
+        {
+            private BitMask[] _data;
```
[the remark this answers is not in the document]
roberthusak: The shortest name sufficiently describing the semantics I could think of was `varToCopySetMap`, which sounds a bit cumbersome. It's the only field in the structure and its semantics is described in the structure summary, so I thought it might be enough. Do you have any suggestions for a better name?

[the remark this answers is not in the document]
jakubmisek: maybe `_state` or `_varState`

src/Peachpie.CodeAnalysis/CodeGen/Graph/BoundExpression.cs
roberthusak: Yes, it does. I added it to the test in 5c474dc73042954fae72ecc299d9042a9f05909c to be sure.

src/Peachpie.CodeAnalysis/FlowAnalysis/ExpressionAnalysis.cs
```diff
@@ -1416,6 +1498,19 @@ void BindParam(PhpParam expected, BoundArgument givenarg)
                     // TODO: Err, variable or field must be passed into byref argument. foo("hello") where function foo(&$x){}
                 }
             }
+            else if (!expected.Type.IsAnyType && givenarg.Value is BoundVariableRef refvar && refvar.Name.IsDirect)
+            {
+                // Even for variables passed by value we may gain information about their type (if we previously had none),
+                // because not complying with the parameter type would have caused throwing a TypeError
+                var local = State.GetLocalHandle(refvar.Name.NameValue);
+                var localType = State.GetLocalType(local);
+                if (localType.IsAnyType && !localType.IsRef &&
```
jakubmisek: why `!localType.IsRef` ?

src/Tests/Peachpie.DiagnosticTests/tests/type_analysis_09.php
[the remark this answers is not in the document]
jakubmisek: actually no; I still don't see if successful call to `foo($a)` implies that `$a` will be an array. The test only shows that it behaves like in PHP which says the behavior is "undefined" :)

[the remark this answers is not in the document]
roberthusak: OK. Well, the main point of the test was to show that anything else than array causes `TypeError` to be thrown, so it must be an array to proceed after the method call. And also as a regression test if PHP changes its mind :-) But it's not bulletproof, that's right.

src/Peachpie.CodeAnalysis/FlowAnalysis/Passes/TransformationRewriter.CopyAnalysis.cs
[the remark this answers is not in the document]
roberthusak: OK, in fc85bbee80464887d9a9d9d8e4f10b5887f079c3.

src/Peachpie.CodeAnalysis/CodeGen/VariableReference.cs
roberthusak: We can add a bitmask to `FlowContext`. That's still not perfect, as the analysis is actually not performed during the main analysis phase, but maybe better than now?

src/Peachpie.CodeAnalysis/FlowAnalysis/ExpressionAnalysis.cs
roberthusak: Good point, fixed in e8f7755671f70fa15bd3f069454a58a54ffb6d0a.

src/Peachpie.CodeAnalysis/FlowAnalysis/ExpressionAnalysis.cs
roberthusak: Good point, fixed in e8f7755671f70fa15bd3f069454a58a54ffb6d0a.

src/Peachpie.CodeAnalysis/FlowAnalysis/ExpressionAnalysis.cs
roberthusak: Yeah, this was a bit unfortunate, fixed in e8f7755671f70fa15bd3f069454a58a54ffb6d0a (just simplified a little).

src/Peachpie.CodeAnalysis/Semantics/Graph/BuilderVisitor.cs
roberthusak: The original `NewBlock()` is still there (and still used on some places, e.g. passed as a delegate to a generator builder), so it would collide with it. I wanted to distinguish this method from it as it doesn't allocate the `Ordinal` number (and the created block is expected to obtain it otherwise).

src/Peachpie.CodeAnalysis/Utilities/BitMask.cs
roberthusak: I knew only about `TypeRefMask`, which is specialized for type references. But as you mentioned it, I found out that there is also `PhpSignatureMask`. Should I use that? From its name and location it seems to be meant for usage in PHP symbols. I haven't found any other type (apart from directly using `ulong`).

src/Peachpie.CodeAnalysis/Utilities/FixpointAnalysis.cs
roberthusak: I wanted it to be understood as a separate lightweight tool, so I didn't put it to `FlowAnalysis` folder next to the classes used in the main analysis phase. But if it makes more sense (it's true that it's dependent on the CFG), I can move it there.

src/Peachpie.Runtime/Operators.cs
roberthusak: There's no such method, should I implement it? Similarly to `IsPhpArray`?

src/Peachpie.Runtime/Operators.cs
roberthusak: Changed to `IsPhpArray`, which handles that, in e8f7755671f70fa15bd3f069454a58a54ffb6d0a.

src/Peachpie.CodeAnalysis/Semantics/Graph/SingleBlockWalker.cs
roberthusak: It was needed to properly analyse each block separately (including the condition in its adjacent edge), without traversing the others, enabling to separate the order of block traversal from the visitor itself. It was inserted in the hierarchy between `GraphWalker` and `GraphVisitor`, dividing `GraphWalker`'s functionality a bit.

src/Peachpie.CodeAnalysis/FlowAnalysis/ExpressionAnalysis.cs
roberthusak:
```php
function foo(array $a, &$b) {
  $b = "foo";
}

function test() {
  $a = [42];
  $b =& $a;
  foo($a, $b);
  echo $a;
}

test();
```
`$a` then contains the string instead of the original array.

src/Peachpie.CodeAnalysis/CodeGen/Graph/BoundExpression.cs
roberthusak: What is `T`? The method needs an `object` as an argument.

src/Peachpie.Runtime/Reflection/PhpRoutineInfo.cs
roberthusak: OK, updated in 2828173f318c841df8a368d1e4817d6ec702e2b6.

src/Peachpie.Runtime/Reflection/PhpRoutineInfo.cs
roberthusak: OK, updated in 2828173f318c841df8a368d1e4817d6ec702e2b6.

src/Peachpie.Runtime/Reflection/PhpRoutineInfo.cs
roberthusak: OK, updated in 2828173f318c841df8a368d1e4817d6ec702e2b6.

src/Peachpie.CodeAnalysis/CodeGen/Graph/BoundExpression.cs
jakubmisek: ok, casting to the `T` (the actual method declaring class) would cause a more readable exception in case type analysis was wrong

src/Peachpie.CodeAnalysis/Symbols/CoreType.cs
```diff
@@ -179,6 +179,7 @@ public CoreTypes(PhpCompilation compilation)
             IPhpArray = Create("IPhpArray");
             IPhpEnumerable = Create("IPhpEnumerable");
             IPhpCallable = Create("IPhpCallable");
+            PhpCallback = Create("PhpCallback");
```
jakubmisek: can be removed if you move the helper method into `Operators`

src/Peachpie.Runtime/Operators.cs
jakubmisek: added in https://github.com/peachpiecompiler/peachpie/commit/bb0116bdfa124b772ea1c2779e3cc8cd91f431dc

src/Peachpie.CodeAnalysis/Semantics/Graph/SingleBlockWalker.cs
jakubmisek: can just `GraphVisitor` traverse like that? or we need two different visitor semantics

src/Peachpie.Runtime/Operators.cs
roberthusak: Thanks, updated in f1b7f783f21a6a8e9439d268e1da33251935a066.

src/Peachpie.CodeAnalysis/Symbols/CoreType.cs
roberthusak: OK, moved in 8615dc96.

src/Peachpie.CodeAnalysis/Semantics/Graph/SingleBlockWalker.cs
roberthusak: We can make `GraphWalker` traverse like that (`GraphVisitor`'s `Visit*` methods are empty). Due to its lack of infinite recursion prevention it cannot be used in a standalone way anyway. We can move its traversal of the edges to `GraphExplorer` (`AnalysisWalker` overrides all the edge visiting anyway). In that case, we might better rename it to something like `BlockWalker`?

jakubmisek: sounds better than fourth visitor class

src/Peachpie.CodeAnalysis/CodeGen/VariableReference.cs
jakubmisek: `SkipPass` only makes sense for `SourceParameter` right ?

roberthusak: Yes.

jakubmisek: good, then it makes more sense to make a property in `SourceParameterSymbol` called i.e. `CopyOnPass` ? will be false for *byref*, and some specific types (*string*, *bool*, *long*, *double*, *object*, ..., even nullable) and when your analysis says that.

src/Peachpie.CodeAnalysis/Utilities/BitMask.cs
jakubmisek: can be unified ... for now, maybe just use `ulong` in the analysis ? is 64 bits enough for this purpose?

src/Peachpie.CodeAnalysis/CodeGen/VariableReference.cs
roberthusak: OK, see 96eb30660bd40b4ca997494b0068443601f14761.

src/Peachpie.CodeAnalysis/CodeGen/VariableReference.cs
roberthusak: Updated in 96eb30660bd40b4ca997494b0068443601f14761.

src/Peachpie.CodeAnalysis/Semantics/Graph/BuilderVisitor.cs
roberthusak: I reverted these changes in `BuilderVisitor` in bd78e6ab60247eb0a3ca96f0157d225d8334139c.

src/Peachpie.CodeAnalysis/Utilities/FixpointAnalysis.cs
roberthusak: Removed in 2de4d15c2434bddddd8037aa257f199b82a3d7db.

src/Peachpie.CodeAnalysis/Semantics/Graph/SingleBlockWalker.cs
roberthusak: Removed in 93776b70682f6c429231968456235de00574b5a6, after basing the two new analyses on `AnalysisWalker`.

src/Peachpie.CodeAnalysis/Utilities/BitMask.cs
roberthusak: Yes, 64 bits are enough for both `ParameterAnalysis` and `CopyAnalysis` (according to running them in WordPress). The purpose of `BitMask` is to encapsulate the manipulation with particular bits and handling overflowing indices (any item with index >= 64 is understood as being in the mask - the same behaviour as in `TypeRefMask`). The code which uses `BitMask` seems clearer to me than if we used `ulong` instead. Do we really want to get rid of it?

src/Peachpie.CodeAnalysis/Symbols/Source/SourceParameterSymbol.cs
```diff
@@ -120,6 +126,10 @@ public SourceParameterSymbol(SourceRoutineSymbol routine, FormalParam syntax, in
                     .BindWholeExpression(syntax.InitValue, BoundAccess.Read)
                     .SingleBoundElement()
                 : null;
+
+            // Only copiable types need to be copied when passed (see CodeGenerator.IsCopiable)
+            var ct = routine.DeclaringCompilation.CoreTypes;
+            CopyOnPass = (Type == ct.PhpValue || Type == ct.PhpString || Type == ct.PhpArray);
```
jakubmisek: @jakubmisek PhpCompilation.IsCopiable( Type )

jakubmisek:
nice! thank you Robert :)

PR #618, opened 2020-01-20
curl cookie `#HttpOnly_` support

DRSDavidSoft:
### Description
This adds the missing `#HttpOnly_` prefix, see [ref](https://github.com/curl/curl/blob/9c1806ae4684ec5ef1aeb39bb9f15cece1c27256/lib/cookie.c#L1466-L1494).


jakubmisek:
awesome! thank you David

PR #619, opened 2020-01-21
[WIP] Cookie Jar/File implementation for CURL

### LINKED ISSUE #615: Request implementation for CURLOPT_COOKIEJAR
### Proposal
When the `CURLOPT_COOKIEJAR` option is set, the following changes should be made:  
  
1. After each request is made, the `CookieContainer`'s contents should be serialized using the `CreateCookieArray()` format, and saved to the disk with file path declared in `CURLOPT_COOKIEJAR`  
  
2. Before each request is made, the contents of `CURLOPT_COOKIEJAR` file should be read and parsed, and put into the `CookieContainer`.

DRSDavidSoft:
### Status
- **Work-in-progress**, please do not merge yet.
- Once this is implemented correctly, it should close #615.

### TODO

- [x] Generate "Netscape-style" cookies
- [x] Parse "Netscape-style" cookies
- [x] `CURLOPT_COOKIEJAR` / `CURLOPT_COOKIEFILE`
- [x] Write cookies to file system
- [x] Read cookies from file system
- [ ] Also, it should parse both of the [cookie data formats](https://curl.haxx.se/libcurl/c/CURLOPT_COOKIEFILE.html):
> The cookie data can be in either the old Netscape / Mozilla cookie data format or just regular HTTP headers (Set-Cookie style) dumped to a file.

### Fixes
- [x] call `PrintCookies()` on `curl_close()`
- [ ] <s>call `PrintCookies()` before `Dispose()`, when and **only if** `curl_close()` was not called</s>
 [(ref.)](https://github.com/peachpiecompiler/peachpie/commit/36dcdf3cceabe8f8c8de3c82b1cb1581859c6e9c#r36945263)
- [ ] support header-style cookies in addition to netscape-style
- [ ] cookie value parse format might not match cURL, more research needed


DRSDavidSoft:
### Couple of questions
- I'm not exactly sure if this is how it should work, but shouldn't `ch.Result` _not_ be `null` in `ProcessResponse()`?
- How do I `join()` a `PhpArray` variable, and create a stream from it?

Here's roughly what I'm aiming for:
```cs
// in ProcessResponse(), why `ch.Result` is null here?
PhpArray cookieJar = (ch.CookieContainer != null) ? CreateCookiePhpArray(ch.Result.Cookies) : PhpArray.Empty;

```

jakubmisek:
Thank you David!

> `ch.Result` is not created yet, it is created by the function `ProcessResponse`

> According to https://curl.haxx.se/libcurl/c/CURLOPT_COOKIEJAR.html cookies should be stored into the specified file upon closing the resource handle, Also the option only allows a `string` value, not streams

How does the netscape-style look like?

DRSDavidSoft:
@jakubmisek 

> `ch.Result` is not created yet, it is created by the function `ProcessResponse`

Oh, I see.  So how can I access the cookies within the `ProcessResponse()` function?

> According to https://curl.haxx.se/libcurl/c/CURLOPT_COOKIEJAR.html cookies should be stored into the specified file upon closing the resource handle

Yes, I didn't realize that -- so the logic I'm trying to implement should actually happen in `curl_close()`, not in `curl_execute()`

Does that mean before the `CURLResource.Dispose()`?

> Also the option only allows a `string` value, not streams

That's correct, it accepts a string which either is a file path, or `-`, which means the generated string should be outputted to `stdout`.

If I'm not mistaken, I should create a file stream that points out to that file path (I'm not sure if this assumption is correct)

> How does the netscape-style look like?

Oh, the `CreateCookiePhpArray()` function actually creates the netscape-style strings, we just need to `join("\n")` / `implode("\n")` it.

Here's an example of the cookie jar file:
```
# Netscape HTTP Cookie File
# https://curl.haxx.se/docs/http-cookies.html
# This file was generated by libcurl! Edit at your own risk.

#HttpOnly_www.domain.com	FALSE	/	FALSE	0	ASP.NET_SessionId	cevhurluif5fijy30slrgrlf
#HttpOnly_www.domain.com	FALSE	/	FALSE	0	__AntiXsrfToken	93d4951dd09f4087a53cd2ce893c06c8
```

---

Additionally, Peach Pie needs a new function to parse these "Netscape-style" cookie files that cURL uses, for when it's set in the `CURLOPT_COOKIEFILE` option.

Example:

```php
curl_setopt($ch, CURLOPT_COOKIEFILE, __DIR__ . '/cookie.txt');
curl_setopt($ch, CURLOPT_COOKIEJAR, __DIR__ . '/cookie.txt' );
```

jakubmisek:
I think you don't have to access them at all; requested cookie container is at `CURLResource.CookieContainer`, and response cookies are in `CURLResource.Result.Cookies` (or if you would need them earlier they come from `HttpWebResponse.Cookies`).

So in `curl_close()` all the cookies are available already;

Parsing netscape-style cookies has to be implemented as well.

jakubmisek:
@DRSDavidSoft I'd like to help - let me draft the structure for option handling

DRSDavidSoft:
@jakubmisek This is actually really helpful! I'll make some comments on your commit

DRSDavidSoft:
@jakubmisek The jar output works at this stage, I'm going to work on the input parsing now.

src/Peachpie.Library.Network/CURLConstants.cs
```diff
@@ -1029,10 +1032,27 @@ public void PrintCookies(Context ctx, CURLResource resource)
...
+                        ""
+                    }));
+
+                    // output the cookies in netscape style:
+                    foreach (var line in CURLFunctions.CookiesToNetscapeStyle(resource.Result?.Cookies))
+                    {
+                        writer.WriteLine(line);
+                    }
+
+                    writer.Close();
```
jakubmisek: I think you don't have to Close(), it gets disposed at the end of `using` block

src/Peachpie.Library.Network/CURLConstants.cs
```diff
@@ -1067,11 +1086,58 @@ public override void Apply(Context ctx, HttpWebRequest request)
...
+
+						// convert expiration date from unix timestamp
+						cookie.Expires = DateTimeOffset.FromUnixTimeSeconds(long.Parse(tokens[4])).UtcDateTime; // TODO: move to DateTimeUtils class?
+
+						// decode cookie name and value
+						cookie.Name = tokens[5];
+						cookie.Value = tokens[6];  // TODO: should HttpUtility.UrlDecode be used here? more research needed
+
+						// add the parsed cookie
+						request.CookieContainer.Add(cookie);
```
DRSDavidSoft: @jakubmisek it seems that this is not working.  Can you please verify/debug this?

Here's the test case:
[**📦 testcase.zip**](https://github.com/peachpiecompiler/peachpie/files/4118394/testcase.zip)

jakubmisek: sorry for the delay, just noticed the comment

DRSDavidSoft: @jakubmisek It's alright. Also I believe I should use a `CookieCollection` first

jakubmisek: `cookie.Expires` is set to `0` which is in Unix world January 1970. 
Hence `cookie.Expired` is true.
Expired cookie is not added to the container but removed instead:

internally:
```c#
			if (cookie.Expired)
			{
				lock (cookieCollection)
				{
					int num = cookieCollection.IndexOf(cookie);
					if (num != -1)
					{
						cookieCollection.RemoveAt(num);
						m_count--;
					}
				}
			}
```

src/Peachpie.Library.Network/CURLConstants.cs
```diff
@@ -1067,11 +1086,58 @@ public override void Apply(Context ctx, HttpWebRequest request)
...
-            // TODO: load netscape-like or header-style cookies from stream
+            // load netscape-like or header-style cookies from stream
 
-            // request.CookieContainer.Add() // add parsed cookie
+            // TODO: add support for header-style cookies (`Set-Cookie: ...`).
+            //       currently, only curl's netscape-like cookies are supported.
 
             // request.CookieContainer.SetCookies( ... ) // set header style cookies
+
+            using (var reader = new StreamReader(stream.RawStream, Encoding.ASCII))
```
jakubmisek: `PhpStream` already has a method `ReadLine` - you don't have to create `StreamReader`

src/Peachpie.Library.Network/CURLConstants.cs
```diff
@@ -1067,11 +1086,58 @@ public override void Apply(Context ctx, HttpWebRequest request)
...
+						// get tokens in an array and trim them
+						string[] tokens = line.Split('\t').Select(token => token.Trim()).ToArray();
+
+						cookie.Domain = tokens[0]; // The domain that created AND can read the variable.
+						// var subdomainAccess = bool.Parse(tokens[1]); // boolean value indicating if all machines within a given domain can access the variable.
+						cookie.Path = tokens[2] ?? "/"; // The path within the domain that the variable is valid for.
+						cookie.Secure = bool.Parse(tokens[3]); // boolean value indicating if a secure connection with the domain is needed to access the variable.
+
+						// convert expiration date from unix timestamp
+						cookie.Expires = DateTimeOffset.FromUnixTimeSeconds(long.Parse(tokens[4])).UtcDateTime; // TODO: move to DateTimeUtils class?
```
jakubmisek: if `tokens[4]` is `0`, the expiration is not specified and time cannot be created

src/Peachpie.Library.Network/CURLConstants.cs
```diff
@@ -1067,11 +1086,58 @@ public override void Apply(Context ctx, HttpWebRequest request)
...
+            using (var reader = new StreamReader(stream.RawStream, Encoding.ASCII))
+            {
+				string line = String.Empty;
+
+				// iterate over the cookie jar lines
+				while ((line = reader.ReadLine()) != null)
+				{
+
+					// gather current cookie information in a new cookie object
+					var cookie = new Cookie();
```
jakubmisek: the line should be parsed first, verified, and then `Cookie` object created

src/Peachpie.Library.Network/CURLFunctions.cs
```diff
@@ -208,7 +208,7 @@ public static PhpValue curl_getinfo(CURLResource ch, int opt = 0)
                 case CURLConstants.CURLINFO_PRIVATE:
                     return r.Private.IsSet ? r.Private.DeepCopy() : PhpValue.False;
                 case CURLConstants.CURLINFO_COOKIELIST:
-                    return ((ch.CookieContainer != null && ch.Result != null) ? CreateCookiePhpArray(ch.Result.Cookies) : PhpArray.Empty);
+                    return CreateCookiePhpArray(ch.Result?.Cookies) ?? PhpArray.Empty;
```
jakubmisek: function never returns `null`, `??` is unecessary

src/Peachpie.Library.Network/CURLFunctions.cs
```diff
@@ -233,7 +233,7 @@ internal static IEnumerable<string> CookiesToNetscapeStyle(CookieCollection cook
 
         static PhpArray CreateCookiePhpArray(CookieCollection cookies)
         {
-            return new PhpArray(CookiesToNetscapeStyle(cookies));
+            return (cookies != null ? new PhpArray(CookiesToNetscapeStyle(cookies)) : PhpArray.Empty);
```
jakubmisek: maybe check `null` in `CookiesToNetscapeStyle` instead?

src/Peachpie.Library.Network/CURLConstants.cs
jakubmisek: We can change `CookiaContainer` to `CookieCollection`. Just note, eventually it will be copied into the CookieContainer anyways (and internally it uses CookieCollection as well)

DRSDavidSoft: @jakubmisek I see!

libcurl writes `0` for expired when the `Expires` in [Set-Cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#Syntax) directive is omitted, e.g. `Set-Cookie: foo=bar; Path=/`.

Thanks for your help; I'll see how this is implemented in libcurl to reflect it here.

jakubmisek: I see, didn't know that! would be interesting to see how to setup the cookie in .NET to work the same

src/Peachpie.Library.Network/CURLConstants.cs
DRSDavidSoft: ### Resolved

```cs
while ((line = stream.ReadLine(-1, null)) != null)
```

src/Peachpie.Library.Network/CURLConstants.cs
DRSDavidSoft: @jakubmisek I'll see how libcurl handles this, I believe an expiration date of `0` means the cookie is a "session cookie".

src/Peachpie.Library.Network/CURLConstants.cs
DRSDavidSoft: Resolved. This was required for the `HttpOnly` flag, so I used a bool instead.

src/Peachpie.Library.Network/CURLConstants.cs
DRSDavidSoft: @jakubmisek I need an example for this. How do read line by line until I reach the end of the file?

jakubmisek: the same `while ((line = stream.ReadLine()) != null)`

jakubmisek:
```c#
// using PhpStream to read lines
while (!stream.Eof) {
   line = stream.ReadLine(-1, null);
}
```

src/Peachpie.Library.Network/CURLConstants.cs
DRSDavidSoft: @jakubmisek  I did a little bit of digging around, and it seems it's similar in .NET's [Cookie.Expires Property Remarks](https://docs.microsoft.com/en-us/dotnet/api/system.net.cookie.expires#remarks) reference:

> Setting the `Expires` property to [DateTime.MinValue](https://docs.microsoft.com/en-us/dotnet/api/system.datetime.minvalue) **makes this a session cookie, which is its _default_ value**.

Since this is the default value for the `Expires` method, I quickly verified that it worked by commenting out the offending line.

So, I used something like this to solve the issue:
```cs
long ts = long.Parse(tokens[4]);

cookie.Expires = ts == 0 ? DateTime.MinValue : DateTimeOffset.FromUnixTimeSeconds(ts).UtcDateTime;

// later moved to DateTimeUtils, so:
// cookie.Expires = DateTimeUtils.UnixTimeStampToUtc(ts);
```

BTW, something like this is used the other way around when generating the netscape-style cookies:

```cs
long expires = (c.Expires.Ticks == 0) ? 0 : DateTimeUtils.UtcToUnixTimeStamp(c.Expires);
```

DRSDavidSoft:
@jakubmisek Could you please review the new changes? I implemented your suggestions, and wrote a fix for the session cookies.

src/Peachpie.Library.Network/CURLConstants.cs
```diff
@@ -1067,11 +1086,71 @@ public override void Apply(Context ctx, HttpWebRequest request)
...
+                {
+                    // get tokens in an array and trim them
+                    string[] tokens = line.Split('\t').Select(token => token.Trim()).ToArray();
+
+                    // gather current cookie information in a new cookie object
+                    var cookie = new Cookie();
+
+                    cookie.Domain = tokens[0].TrimStart('.'); // The domain that created AND can read the variable.
+                    // var subdomainAccess = bool.Parse(tokens[1]); // boolean value indicating if all machines within a given domain can access the variable.
+                    cookie.Path = tokens[2] ?? "/"; // The path within the domain that the variable is valid for.
```
jakubmisek: tokens[2] is never `null`
so `??` does not make sense

src/Peachpie.Library.Network/CURLConstants.cs
```diff
@@ -1067,11 +1086,71 @@ public override void Apply(Context ctx, HttpWebRequest request)
...
+                    cookie.Secure = bool.Parse(tokens[3]); // boolean value indicating if a secure connection with the domain is needed to access the variable.
+
+                    // normalize cookie path according to curl behavior
+                    cookie.Path = cookie.Path.TrimEnd('/') + '/';
```
jakubmisek: - should be `"/"` (string), `/` causes unnecessary allocation and conversion to string
- recommending to use `PathUtils.TrimEndSeparator(cookie.Path)`, faster without additional allocations

src/Peachpie.Library.Network/CURLConstants.cs
jakubmisek: `??` is unnecessary, line `1136` trims it anyways

src/Peachpie.Library.Network/CURLConstants.cs
```diff
@@ -1067,11 +1086,71 @@ public override void Apply(Context ctx, HttpWebRequest request)
...
+
+                    // convert expiration date from unix timestamp
+                    long expires = long.Parse(tokens[4]);
```
jakubmisek: maybe use `long.TryParse` instead ? this might cause an exception

src/Peachpie.Library.Network/CURLConstants.cs
```diff
@@ -1067,11 +1086,71 @@ public override void Apply(Context ctx, HttpWebRequest request)
...
+                    // get tokens in an array and trim them
+                    string[] tokens = line.Split('\t').Select(token => token.Trim()).ToArray();
+
+                    // gather current cookie information in a new cookie object
+                    var cookie = new Cookie();
+
+                    cookie.Domain = tokens[0].TrimStart('.'); // The domain that created AND can read the variable.
+                    // var subdomainAccess = bool.Parse(tokens[1]); // boolean value indicating if all machines within a given domain can access the variable.
+                    cookie.Path = tokens[2] ?? "/"; // The path within the domain that the variable is valid for.
+                    cookie.Secure = bool.Parse(tokens[3]); // boolean value indicating if a secure connection with the domain is needed to access the variable.
```
jakubmisek: maybe use `bool.TryParse` instead ? this might cause an exception

src/Peachpie.Runtime/Utilities/DateTimeUtils.cs
```diff
@@ -34,5 +34,15 @@ public static double UtcToUnixTimeStampFloat(DateTime dt)
...
         }
+
+        /// <summary>
+        /// Converts UNIX timestamp to <see cref="DateTime"/> representing UTC time.
+        /// </summary>
+        /// <param name="ts">Unix timestamp.</param>
+        /// <returns>Time.</returns>
+        public static DateTime UnixTimeStampToUtc(long ts)
+        {
+            return ts == 0 ? DateTime.MinValue : DateTimeOffset.FromUnixTimeSeconds(ts).UtcDateTime;
```
jakubmisek: it's not true, if `0` has a special meaning in cookies, please handle it in cookies, not in Runtime Utilities

src/Peachpie.Library.Network/CURLConstants.cs
DRSDavidSoft: @jakubmisek I believe `PathUtils.TrimEndSeparator()` implementation will only remove the last present `/` character, whereas `TrimEnd()` will remove as many as `/` characters present:
https://github.com/peachpiecompiler/peachpie/blob/6e8d6649a3110c6f6b640ad862515e699d5f214b/src/Peachpie.Runtime/Utilities/PathUtils.cs#L28-L33
Should this behavior be changed? Consider the following string:

```cs
string somePath = "/some/path///"
```

jakubmisek: you are right, so please keep `TrimEnd` (if it's what curl does)

DRSDavidSoft: @jakubmisek I'll stick to `TrimEndSeparator()` since cookie behaves [the same](https://github.com/curl/curl/blob/9c1806ae4684ec5ef1aeb39bb9f15cece1c27256/lib/cookie.c#L326-L329) way.

src/Peachpie.Library.Network/CURLFunctions.cs
DRSDavidSoft: @jakubmisek I'd appreciate it if you could edit this pull request to reflect the changes for this one

DRSDavidSoft:
@jakubmisek Can you have another look, please?

src/Peachpie.Library.Network/CURLConstants.cs
```diff
@@ -1067,11 +1086,88 @@ public override void Apply(Context ctx, HttpWebRequest request)
...
+
+            // iterate over the cookie jar lines
+            while (!stream.Eof)
+            {
+
+                // read a line, and strip off the possible end-of-line characters
+                line = stream.ReadLine(-1, null).Trim();
+
+                // ignore blank lines
+                if (String.IsNullOrWhiteSpace(line))
```
jakubmisek: won't be a whitespace :) can be changed to `IsNullOrEmpty()`

src/Peachpie.Library.Network/CURLConstants.cs
```diff
@@ -1067,11 +1086,88 @@ public override void Apply(Context ctx, HttpWebRequest request)
...
+                    continue;
+
+                bool httpOnly;
+
+                // check for HttpOnly cookies and remove the #HttpOnly prefix
+                if (line.Length < 10 || line.Substring(0, 10) != "#HttpOnly_")
```
jakubmisek: `line.StartsWith( ... )` please

jakubmisek:
thank you @DRSDavidSoft !

src/Peachpie.Library.Network/CURLConstants.cs
DRSDavidSoft: @jakubmisek Force of habit :D Even though I `.Trim()` it just one line above

PR #649, opened 2020-02-01
Implemented DOMElement::getElementById()

### LINKED ISSUE #648: DOMDocument->getElementById() NotImplementedException
### Description
The `DOMElement::getElementById()` method seems not to be implemented:

https://github.com/peachpiecompiler/peachpie/blob/55ab289c1f438237c93825c5d6e16ca2a3ad5370/src/Peachpie.Library.XmlDom/DOMDocument.cs#L483-L489

Could you please implement it?

DRSDavidSoft:
### Scope
- fixes #648

### DIFF
--- a/src/Peachpie.Library.XmlDom/DOMDocument.cs
+++ b/src/Peachpie.Library.XmlDom/DOMDocument.cs
@@ -481,11 +481,14 @@ namespace Peachpie.Library.XmlDom
         }
 
         /// <summary>
-        /// Not yet implemented.
+        /// Gets the first element with the matching ID attribute.
         /// </summary>
+        /// <param name="elementId">The attribute ID to match.</param>
+        /// <returns>A <see cref="DOMElement"/>.</returns>
         public virtual DOMElement getElementById(string elementId)
         {
-            throw new NotImplementedException();
+            XmlElement element = XmlDocument.GetElementById(elementId);
+            return new DOMElement(element);
         }
 
         #endregion
--- a/src/Peachpie.Library.XmlDom/DOMDocument.cs
+++ b/src/Peachpie.Library.XmlDom/DOMDocument.cs
@@ -488,7 +488,7 @@ namespace Peachpie.Library.XmlDom
         public virtual DOMElement getElementById(string elementId)
         {
             XmlElement element = XmlDocument.GetElementById(elementId);
-            return new DOMElement(element);
+            return element != null ? new DOMElement(element) : null;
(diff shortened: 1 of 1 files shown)


DRSDavidSoft:
It's funny, I don't see why this hasn't been not implemented before 😄

src/Peachpie.Library.XmlDom/DOMDocument.cs
```diff
@@ -481,11 +481,14 @@ public virtual DOMNodeList getElementsByTagNameNS(string namespaceUri, string lo
...
-        /// Not yet implemented.
+        /// Gets the first element with the matching ID attribute.
         /// </summary>
+        /// <param name="elementId">The attribute ID to match.</param>
+        /// <returns>A <see cref="DOMElement"/>.</returns>
         public virtual DOMElement getElementById(string elementId)
         {
-            throw new NotImplementedException();
+            XmlElement element = XmlDocument.GetElementById(elementId);
+            return new DOMElement(element);
```
jakubmisek: the function `getElementById` returns `null` if the element is not found

jakubmisek:
Thanks! the [documentation](http://docs.php.net/manual/en/domdocument.getelementbyid.php) is little bit more complex, but I think .NET handles it well :)

src/Peachpie.Library.XmlDom/DOMDocument.cs
DRSDavidSoft: @jakubmisek So, `DOMElement(null)` does not return `null`?

jakubmisek: `new DOMElement` certainly does not :)

DRSDavidSoft: @jakubmisek Which one would you suggest is the better style?

### A
```cs
return element == null ? element : new DOMElement(element);
```

### B
```cs
return element == null ? null : new DOMElement(element);
```

### C
```cs
return element != null ? new DOMElement(element) : null;
```

jakubmisek: haha :) `C` is fine - the most frequent branch first

DRSDavidSoft: @jakubmisek Awesome, never thought of it that way.
I just committed this for review

jakubmisek:
thanks :)

Issue #402, opened 2019-04-02
Laravel: OpenSSL extension seems to be missing


calvinbaart:
Laravel uses the OpenSSL extension for its Encryption facilities.

To work around this I needed to stub them in a dummy file:
```php
function openssl_cipher_iv_length(string $method)
{
    return 32;
}

function openssl_encrypt(string $data, string $method, string $key, int $options = 0, string $iv = "", string &$tag = null, string $aad = "", int $tag_length = 16): string
{
    return $data;
}

function openssl_decrypt(string $data, string $method, string $key, int $options = 0, string $iv = "", string $tag = "", string $aad = ""): string
{
    return $data;
}
```

jakubmisek:
implemented in https://github.com/peachpiecompiler/peachpie/pull/688, closing

PR #688, opened 2020-02-26
Feature/openssl

jakubmisek:
implements https://github.com/peachpiecompiler/peachpie/issues/402


kripper:
How can openssl be included?
I tried adding:
  <ItemGroup>
	<PackageReference Include="Peachpie.Library.OpenSSL"/>
  </ItemGroup>

jakubmisek:
it's referenced by default as a part of `Peachpie.Library`

kripper:
We are using:
```
  <ItemGroup>
	<PackageReference Include="Peachpie.Library"/>
  </ItemGroup>
```
But we get: `warning PHP5006: Call to undefined function: 'openssl_sign'` with Version=0.9.981.0

jakubmisek:
literally; just read my reply above

Issue #700, opened 2020-03-11
support for int64 array keys


jakubmisek:
```php
$key = PHP_INT_MAX;

$arr = [$key => $key]; // OverflowException
$arr = [PHP_INT_MAX => $key]; // [-1]
```

PHP_INT_MAX is used as an array key sometimes which is correct.

PeachPie internally uses int32 as the array key resulting to two different behaviors:
- when passed in compile-time, the int64 number is casted silently to int32
- when passed in run-time, the OverflowException is thrown 

Note: internally `IntStringKey` and PhpArray have to support int64 indexes

PR #701, opened 2020-03-11
implements 64 bit array keys

jakubmisek:
fixes https://github.com/peachpiecompiler/peachpie/issues/700

`PhpArray` (representing PHP's `array` type) now works with `Int64` keys.


jakubmisek:
thanks Robert! Good one, I think this case already didn't work before the PR so I'll make an issue for it :) 

If you'd have any idea about improving the performance of `OrderedDictionary`, I'm listening :))

PR #719, opened 2020-04-02
Fix cookie handling in UserHandlerInternal session handler

### LINKED ISSUE #718: Problematic handling of session cookie when custom session handler is used
As a test case, let's have the following as the single PHP script in an Asp.NET Core website (`FileSessionHandler` is taken from the [PHP manual](https://www.php.net/manual/en/function.session-set-save-handler.php)):
```php
<?php

class FileSessionHandler implements SessionHandlerInterface
{
  private $savePath;

  function open($savePath, $sessionName)
  {
    $this->savePath = $savePath;
    if (!is_dir($this->savePath)) {
      mkdir($this->savePath, 0777);
    }

    return true;
  }

  function close()
  {
    return true;
  }

  function read($id)
  {
    return (string)@file_get_contents("$this->savePath/sess_$id");
  }

  function write($id, $data)
  {
    return file_put_contents("$this->savePath/sess_$id", $data) === false ? false : true;
  }

  function destroy($id)
  {
    $file = "$this->savePath/sess_$id";
    if (file_exists($file)) {
      unlink($file);
    }

    return true;
  }

  function gc($maxlifetime)
  {
    foreach (glob("$this->savePath/sess_*") as $file) {
      if (filemtime($file) + $maxlifetime < time() && file_exists($file)) {
        unlink($file);
      }
    }

    return true;
  }
}

session_set_save_handler(new FileSessionHandler(), true);
session_start();
echo session_id() ."\n";
```

Expected behavior: The session ID stays the same between subsequent page reloads and is stored as a cookie.

Actual behavior: The session ID is different for each page load, not stored in cookies and the following exception is thrown due to setting the cookie from the call of `session_write_close` (registered as a shutdown handler by `session_set_save_handler`):
```
System.InvalidOperationException: Headers are read-only, response has already started.
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpHeaders.ThrowHeadersReadOnlyException()
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpHeaders.Microsoft.AspNetCore.Http.IHeaderDictionary.set_Item(String key, StringValues value)
   at Microsoft.AspNetCore.Http.ResponseCookies.Append(String key, String value, CookieOptions options)
   at Peachpie.AspNetCore.Web.RequestContextCore.Pchp.Core.IHttpPhpContext.AddCookie(String name, String value, Nullable`1 expires, String path, String domain, Boolean secure, Boolean httpOnly)
   at Pchp.Library.Session.UserHandlerInternal.Persist(IHttpPhpContext webctx, PhpArray session)
   at Pchp.Core.PhpSessionHandler.CloseSession(Context ctx, IHttpPhpContext webctx, Boolean abandon)
   at Pchp.Library.Session.session_commit(Context ctx)
   at Pchp.Library.Session.session_write_close(Context ctx)
   at Pchp.Core.Context.ProcessShutdownCallbacks()
   at Pchp.Core.Context.Dispose()
   at Peachpie.AspNetCore.Web.RequestContextCore.Dispose()
   at Peachpie.AspNetCore.Web.PhpHandlerMiddleware.Invoke(HttpContext context)
   at Microsoft.AspNetCore.Session.SessionMiddleware.Invoke(HttpContext context)
   at Microsoft.AspNetCore.Session.SessionMiddleware.Invoke(HttpContext context)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol.ProcessRequests[TContext](IHttpApplication`1 application)
```

roberthusak:
Fixes #718

### DIFF
--- a/src/Peachpie.Library/Session.cs
+++ b/src/Peachpie.Library/Session.cs
@@ -295,7 +295,10 @@ namespace Pchp.Library
                     return null;
                 }
 
-                // 3. read
+                // 3. set session cookie
+                TrySetCookie(webctx, warnIfHeadersSent: true);
+
+                // 4. read
                 var str = _handler.read(GetSessionId(webctx));
                 if (str.IsEmpty)
                 {
@@ -308,7 +311,7 @@ namespace Pchp.Library
 
             public override bool Persist(IHttpPhpContext webctx, PhpArray session)
             {
-                webctx.AddCookie(GetSessionName(webctx), _lazyid, null); // TODO: lifespan
+                TrySetCookie(webctx, warnIfHeadersSent: false);
 
                 //
                 var handler = GetSerializeHandler((Context)webctx);
@@ -325,7 +328,7 @@ namespace Pchp.Library
             {
                 if (!_isnewsession)
                 {
-                    webctx.AddCookie(GetSessionName(webctx), string.Empty, DateTimeOffset.UtcNow);
+                    TryUnsetCookie(webctx);
                 }
 
                 // destroy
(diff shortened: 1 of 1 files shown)


jakubmisek:
This should be done in `UserHandlerInternal.GetSessionId()` which supposes to get or lazily create session object and eventually set the corresponding cookie. If it cannot set the cookie, it must throw the original an exception.

(ref aspnet: System.Web.SessionState.SessionStateModule.DelayedGetSessionId())

"TrySetCookie" or be silent does not seem to be a good way of handling user data.

PR #740, opened 2020-05-02
Using a Task and TaskCompletionSource to give control to RequestContextCore

oferns:
The idea behind this pr is that we pass a TaskCompletionSource to the RequestContextCore.ProcessScript method so that it can decide when to end the response. It does this by setting a private field which can be accessed by other members  (ie Flush, which is called by fastcgi_finish_request) or by calling Task.SetResult or Task.SetException. The TaskCompletionSource.Task is returned by the middleware, and the execution of RequestContextCore.ProcessScript is done in a non-awaited task. It seems to work. Its a bit PoC..are there any issues running the ProcessScript in a task? Also, should we call ConfigureAwait(false) on the Task.Run in the Invoke method of the middleware?

### DIFF
--- a/src/Peachpie.AspNetCore.Web/PhpHandlerMiddleware.cs
+++ b/src/Peachpie.AspNetCore.Web/PhpHandlerMiddleware.cs
@@ -89,16 +89,17 @@ namespace Peachpie.AspNetCore.Web
 
         public Task Invoke(HttpContext context)
         {
+            var taskSource = new TaskCompletionSource<object>();
             var script = RequestContextCore.ResolveScript(context.Request);
             if (script.IsValid)
             {
                 using (var phpctx = new RequestContextCore(context, _rootPath, _options.StringEncoding))
                 {
-                    OnContextCreated(phpctx);
-                    phpctx.ProcessScript(script);
+                    OnContextCreated(phpctx);                    
+                    Task.Run(() => phpctx.ProcessScript(script, taskSource)); // fire & forget
+                    return taskSource.Task; // At somepoint we need to set the taskSource result/exception
                 }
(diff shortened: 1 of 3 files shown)


jakubmisek:
Good, I get the idea. ProcessScript must run on the same thread as Context, it is not thread-safe, also there will be some issues with response being disposed while script is still running.

Next this would have to be optimized, we try to make it allocation free, and as performant as possible, this adds a lot of additional overhead just because of a single platform dependant function :/ 

Wouldn't AutoResetEvent work just fine?

oferns:
I dont think we can use ManualResetEvent/AutoResetEvent because we still have to wait for ProcessScript to return before we can return the task that has the AutoResetEvent.Wait call in it. I have created a separate PR/Branch to show what I mean

jakubmisek:
FInalizeBuffer is called as a part of Dispose, which should not be called twice.

This seems like too much logic just because of a minor feature, let me try to make it simpler (it should be a single something, combined with request timeout feature, plus handled in response output stream which would throw exceptions in this case)

jakubmisek:
This was the idea, timeout and "end" implemented as a single `ManualResetEvent` without too much unnecessary code around

https://github.com/peachpiecompiler/peachpie/commit/4b0541a1d2b2b749e666f87afb29e3595cdb1228

oferns:
I will pull this and let you know what I find. I see what you are getting at. It might not be until tomorrow now. Your time is much appreciated. Thank you.

jakubmisek:
thanks! I've made changes in https://github.com/peachpiecompiler/peachpie/pull/745 using ManualResetEvent and merged

PR #741, opened 2020-05-03
Feature/fastcgi finish request alt

oferns:
This an investigation into using the a Reset Event to end the http response early. It doesnt do what we want as it still has to wait for ProcessScript to finish before returning the Task. Its here for discussion only.

### DIFF
--- a/src/Peachpie.AspNetCore.Web/PhpHandlerMiddleware.cs
+++ b/src/Peachpie.AspNetCore.Web/PhpHandlerMiddleware.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.IO;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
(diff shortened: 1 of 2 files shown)


jakubmisek:
`ProcessScript` would have run on thread

hez2010:
Seems that it better to make `script.Evaluate()` async.

jakubmisek:
CancellationToken would not work unless we would be fully async or handle CancellationToken.IsCancelRequested explicitly

Making Evaluate async is HUGE feature that breaks everything and causes an extreme overhead to all the script inclusion.

hez2010:
How about using `ValueTask` while making `Evaluate` async? It's almost the same with `Task` but not going to allocate on heap, which is 0 cost.

jakubmisek:
we are not going to change the entire inclusion logic to async now :) not relevant to this PR

jakubmisek:
thanks! I've summarized these changes in https://github.com/peachpiecompiler/peachpie/pull/745 and merged

Issue #772, opened 2020-06-25
Feature Request:  DOMNode::C14N() support


kripper:
Feature Request:  DOMNode::C14N() support

If necessary, we can provide a simple test case and extensive testing.

kripper:
Implemented here: https://github.com/kripper/peachpie/commit/79c054f89a7e96e4430bd3a5ee2685aeeddf3e85

jakubmisek:
thanks, can we have a pull request for this? without the other changes and reverted changes

kripper:
Here: https://github.com/peachpiecompiler/peachpie/pull/783

PR #783, opened 2020-06-30
XML improvements

kripper:
fixes https://github.com/peachpiecompiler/peachpie/issues/772


jakubmisek:
The test case for `asXml()` gives me XML declaration in both PHP 7.4 and PeachPie.

Isn't there some additional configuration you have?

kripper:
> The test case for `asXml()` gives me XML declaration in both PHP 7.4 and PeachPie.
> http://sandbox.onlinephpfunctions.com/code/fb25b0e262a143496e9a666dd89b52ce4fd741cc

Right. Provided test case was wrong, because it was returning the root element where it's ok to include XML tag. I should have provided another test case for returning a child element. Anyway, I guess you already figured out and merged. Thanks.

PR #796, opened 2020-07-22
Implementation of remaining methods in xmlwriter

TomatorCZ:
There are a few differences.
Php's indentation is a little bit different and it would be difficult to simulate it in C#. I thing it is not a big deal deal because an indentation is almost for people.
Dtd declaration can be only placed in prolog due to XML specification, but Php tolerates it after closed root element. I thing this corner case doesn't appear because of well-formed xml.
Setting indentation can be done only before the xmlwriter starts writting in C#. I thing there aren't situations, when you change a settings among the tags.
Php allows to write elements into comment, cdata sections. Is is simulated in C# as well, but some corner cases behave in a little different way. e.g. when you don't close an element until comment section end, C# closes it. Is is a weird feature of Php and i thing it is useless.


jakubmisek:
Thanks! May I ask to merge changes from master into this PR? There are some commits related to SmpleXml::asXML and XmlDom

jakubmisek:
a quick question - does it have to be `IDisposable`?

TomatorCZ:
> a quick question - does it have to be `IDisposable`?

If it is not disposed until a script is ended, there might be a problem with handlers (e.g. file handler) held by xmlwriter. I am not sure if all streams are registered in Context as disposable somehow, so to be sure I implemented IDisposable.

PR #800, opened 2020-07-23
Implementation of SimpleXMLIterator

TomatorCZ:
Fixes:
- There was problem with construing derived class from SimpleXMLWriter. See line 425 and test simpleXMLIterator_001 and 005
- Added two parameters in __construct


jakubmisek:
thank you Tomas! will be merged, I'll take a look on `className`, we can do it more lazily I guess, maybe remember `PhpTypeInfo` instead of simple `string`

PR #822, opened 2020-08-30
Fix PDO::FETCH_CLASS with FetchAll()

Indigo744:
`PDO::FETCH_CLASS` was already managed, but not when directly used in `fetchAll()`:
```php
$stmt->fetchAll(\PDO::FETCH_CLASS, "Foo", array("bar"));
```

This PR adds a new test file to check both usage and add a fix for `fetchAll()`:

### DIFF
--- a/src/PDO/Peachpie.Library.PDO/PDOStatement.cs
+++ b/src/PDO/Peachpie.Library.PDO/PDOStatement.cs
@@ -446,7 +446,10 @@ namespace Peachpie.Library.PDO
 
             if ((style & PDO_FETCH.FETCH_CLASS) != 0 && !fetch_argument.IsEmpty)
             {
-                setFetchMode(fetch_style, fetch_argument, ctor_args);
+                if (!setFetchMode(fetch_style, fetch_argument, ctor_args))
+                {
+                    return null;
+                }
             }
 
             var result = new PhpArray();
--- a/src/PDO/Peachpie.Library.PDO/PDOStatement.cs
+++ b/src/PDO/Peachpie.Library.PDO/PDOStatement.cs
@@ -443,7 +443,7 @@ namespace Peachpie.Library.PDO
                 }
             }
 
-            if (fetch_style.HasFlag(PDO_FETCH.FETCH_CLASS) && !fetch_argument.IsEmpty)
+            if ((style & PDO_FETCH.FETCH_CLASS) != 0 && !fetch_argument.IsEmpty)
             {
                 setFetchMode(fetch_style, fetch_argument, ctor_args);
             }
--- a/src/PDO/Peachpie.Library.PDO/PDOStatement.cs
+++ b/src/PDO/Peachpie.Library.PDO/PDOStatement.cs
@@ -428,85 +428,68 @@ namespace Peachpie.Library.PDO
         [return: CastToFalse]
         public virtual PhpArray fetchAll(PDO_FETCH fetch_style = default, PhpValue fetch_argument = default, PhpArray ctor_args = null)
         {
-            // Save internal value that could be overriden by setFetchMode
-            var initial_fetch_column = _fetch_column;
-            var initial_default_fetch_type = _default_fetch_type;
-            var initial_default_fetch_class = _default_fetch_class;
-            var initial_default_fetch_class_args = _default_fetch_class_args;
+            var style = fetch_style != PDO_FETCH.Default ? fetch_style : _default_fetch_type;
+            var flags = style & PDO_FETCH.Flags;
 
-            try
+            if (style == PDO_FETCH.FETCH_COLUMN)
             {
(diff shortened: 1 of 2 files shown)


src/PDO/Peachpie.Library.PDO/PDOStatement.cs
```diff
@@ -443,6 +443,11 @@ public virtual PhpArray fetchAll(PDO_FETCH fetch_style = default, PhpValue fetch
                 }
             }
 
+            if (fetch_style.HasFlag(PDO_FETCH.FETCH_CLASS) && !fetch_argument.IsEmpty)
```
jakubmisek, line 446: Please don't use `Enum.HasFlag()`, it's significantly slow.

Following is recommended:
```C#
(fetch_style & PDO_FETCH.FETCH_CLASS) != 0
```

src/PDO/Peachpie.Library.PDO/PDOStatement.cs
```diff
@@ -443,6 +443,11 @@ public virtual PhpArray fetchAll(PDO_FETCH fetch_style = default, PhpValue fetch
...
+            {
+                setFetchMode(fetch_style, fetch_argument, ctor_args);
```
jakubmisek, line 448: Result of `setFetchMode` should be handled,

src/PDO/Peachpie.Library.PDO/PDOStatement.cs
jakubmisek, line 446: Instead of `fetch_style`, we should probably use `style` (below) since it might fallback to `_default_fetch_type` (previously set default)

The same would apply for `FETCH_COLUMN` above. Or does PHP respect the default value in this case at all?

jakubmisek, line 446: we should not check for `fetch_argument.IsEmpty` here, `setFetchMode` should be called in any case once `fetch_style` is `FETCH_CLASS`

jakubmisek (changes requested):
I have a few notes to the PR :)

jakubmisek:
I'm not sure if it's alright to change `_default_fetch_class` by calling `fetchAll(PDO::FETCH_CLASS, "classname")` .. subsequent calls to `fetchAll()` will use the "classname" instead of previous set `_default_fetch_class`
```php
$pdo->setFetchMode(PDO::FETCH_CLASS, "class1");
$pdo->fetchAll(PDO::FETCH_CLASS, "class2");
$pdo->fetchAll(PDO::FETCH_CLASS); // is it suppose to use "class1" or "class2" ?
```

src/PDO/Peachpie.Library.PDO/PDOStatement.cs
Indigo744: Noted.

Indigo744:
I will add more test case 😸

src/PDO/Peachpie.Library.PDO/PDOStatement.cs
Indigo744: By default value, I guess you mean if there are any differences between:
```php
$pdo->fetchAll(0);
// vs.
$pdo->fetchAll();
```

Indigo744: There are absolutely no differences between these two. PHP must think `0` is `falsy`...

Indigo744: I will move the `FETCH_COLUMN` test to use `style`

jakubmisek:
The last thing is the following test case. `setFetchMode(\PDO::FETCH_CLASS, $className)` should not be overriden by calling `fetchAll(\PDO::FETCH_CLASS, $className2)`.

```php
$stmt->execute();
$stmt->setFetchMode(\PDO::FETCH_CLASS, $className, array(74));
print_r($stmt->fetchAll(\PDO::FETCH_CLASS, $className2)); // uses "className2"
$stmt->execute();
print_r($stmt->fetchAll()); // uses "PDO::FETCH_CLASS" with "className"
```

jakubmisek (approved):

Indigo744:
> The last thing is the following test case. `setFetchMode(\PDO::FETCH_CLASS, $className)` should not be overriden by calling `fetchAll(\PDO::FETCH_CLASS, $className2)`.


@jakubmisek I tried a naive approach: save initial internal values and set them again in a `finally` block.

While it works, I'm not sure I like it...

jakubmisek:
@Indigo744 maybe for now, revert the https://github.com/peachpiecompiler/peachpie/pull/822/commits/5ab2bcc7d9d8021c85a7795067b4f1de7a3286dc please. 

We can do it as a separate issue/PR, it goes far beyond this pull request :)

Indigo744:
@jakubmisek alright! Done 😺

PR #823, opened 2020-08-30
Fix key ordering when fetching using both numeric and associative keys

Indigo744:
Fix the order in which the keys are added to the array to match PHP implementation.


jakubmisek:
thanks!

PR #824, opened 2020-08-30
Implements FETCH_GROUP

### LINKED ISSUE #817: PDO::FETCH_GROUP is not taken into account
Hello,

We use extensively the option [`PDO::FETCH_GROUP`](https://phpdelusions.net/pdo/fetch_modes#FETCH_GROUP) in our code base.

But it seems that it is not supported:
```php
$a = $statement->fetchAll( PDO::FETCH_ASSOC );
$b = $statement->fetchAll( PDO::FETCH_GROUP | PDO::FETCH_ASSOC);
```

Expected:
```
$a = array(1) {
  [0]=> array(32) { ... }
}
$b = array(1) {
  [1] => array(1) {
    [0] => array(31) { ... }
  }
}
```

Instead, `$b` is the same as `$a`.

The thing is that:
1. there is no warning
1. the [compat matrix](https://docs.peachpie.io/compatibility-status/) doesn't show this constant

I know this lib is a work in progress, but there should be a reliable way to tell what is actually implemented, and what isn't.

Thank you.

Indigo744:
Fix #817 

Implements FETCH_GROUP and works with FETCH_ASSOC, FETCH_NUM and FETCH_BOTH.

Note that it only make sense to implement in `fetchAll`, as `fetch` only returns one row (but it is checked in the test case to be sure).

### DIFF
--- a/src/PDO/Peachpie.Library.PDO/PDOStatement.cs
+++ b/src/PDO/Peachpie.Library.PDO/PDOStatement.cs
@@ -470,6 +470,7 @@ namespace Peachpie.Library.PDO
                     break;
 
                 default:
+                    var hasGroupFlag = (style & PDO_FETCH.FETCH_GROUP) != 0;
 
                     for (; ; )
                     {
@@ -477,7 +478,22 @@ namespace Peachpie.Library.PDO
                         if (value.IsFalse)
                             break;
 
-                        result.Add(value);
+                        // Handle FETCH_GROUP case 
+                        if (hasGroupFlag)
+                        {
+                            var row = value.ToArray();
+
+                            // We remove the first column and use it to group rows
+                            var firstCol = row.RemoveFirst().Value;
+
+                            if (!result.ContainsKey(firstCol))
+                                result.Add(firstCol, new PhpArray());
+
+                            result[firstCol].ToArray().Add(row);
+                        } else
+                        {
+                            result.Add(value);
+                        }
                     }
 
                     break;
--- a/src/PDO/Peachpie.Library.PDO/PDOStatement.cs
+++ b/src/PDO/Peachpie.Library.PDO/PDOStatement.cs
@@ -485,6 +485,7 @@ namespace Peachpie.Library.PDO
 
                             // We remove the first column and use it to group rows
                             var firstCol = row.RemoveFirst().Value;
+                            row.ReindexIntegers(0);
 
                             if (!result.ContainsKey(firstCol))
                                 result.Add(firstCol, new PhpArray());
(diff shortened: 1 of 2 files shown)


jakubmisek:
thanks for this :) I wasn't sure how to approach this weird PDO flag.

Although the code seems a bit unclean :)) and internally each of those array operations is very expensive also. Let me think about it.

jakubmisek:
I think FETCH_GROUP is the flag (determined above `var flags = style & PDO_FETCH.Flags`) that should be handled separately after the `result` is populated. So the current switch code block stays intact.

```c#
if (flags == PDO_FETCH.FETCH_GROUP) {
 // transform result array into a new result array
}
```

Indigo744:
I moved the code outside the switch, as requested. Cleaner indeed.

I changed the way numeric keys are determined. I actually check the first row to see if there is a numeric key. I think it's better than checking all flags, since it'll work even if a new flag is added.

In any case, using FETCH_GROUP will always be more expansive than without, but will still be faster than doing it in PHP!

jakubmisek:
good! May I ask for write access to your `FETCH_GROUP` branch? I'd add some cleanups

Indigo744:
> May I ask for write access to your `FETCH_GROUP` branch? I'd add some cleanups

I'm not sure how? I can see _Allow edits by maintainers_ and it is already checked so maybe that's it?

jakubmisek:
> I'm not sure how? I can see _Allow edits by maintainers_ and it is already checked so maybe that's it?

I'm probably not listed within your project *maintainers* group

Indigo744:
Well, that shouldn't be needed, as the docs says about _Allow edits by maintainers_:
> If checked, users with write access to peachpiecompiler/peachpie can add new commits to your FETCH_GROUP branch. You can always change this setting later.

So that should be good.

jakubmisek:
@Indigo744 thanks :)

src/PDO/Peachpie.Library.PDO/PDOStatement.cs
```diff
@@ -485,40 +485,63 @@ public virtual PhpArray fetchAll(PDO_FETCH fetch_style = default, PhpValue fetch
...
-                    // => Always remove the 0 key (FETCH_NUM: does nothing, FETCH_BOTH: remove remaining)
-                    // => Reindex starting from 0
-                    if (hasNumKeys)
-                    {
-                        row.RemoveKey((IntStringKey)0);
-                        row.ReindexIntegers(0);
-                    }
+            // hasNumKeys: when FETCH_NUM or FETCH_BOTH
+            var hasNumKeys =
+                (style & ~PDO_FETCH.Flags) == PDO_FETCH.FETCH_NUM ||
```
Indigo744, line 509: Beware for hasNumKeys, you need to also check for `FETCH_DEFAULT` since FETCH_BOTH is the default...

It was partly why I choose to go to the "check if there is a numeric key in the first row" route...

jakubmisek: good point!

PR #827, opened 2020-08-31
Improve parse_ini implementation

### LINKED ISSUE #826: INI: Value 'Raw' of argument 'scanner_mode' is not supported
Hello,

We have a lot of [`parse_ini_file()` call with `INI_SCANNER_RAW`](https://www.php.net/manual/en/function.parse-ini-string.php), which shows the following warning: 
```
PHP (Warning): Value 'Raw' of argument 'scanner_mode' is not supported
```

The culprit is found here: https://github.com/peachpiecompiler/peachpie/blob/master/src/Peachpie.Library/FileSystem.Ini.cs#L51

What would be the correct approach to handle this flag?

Indigo744:
This is a WIP, do not merge yet.

This PR tracks the work to bring parse_ini closer to PHP implementation:
- [X] Fix truthy/falsy handling
- [x] Fix quoted string with apostrophe (`key='test'`)
- [x] Fix quoted string having escaped quotes
- [x] Fix quoted string multiline
- [x] Fix key with spaces (`foo bar=works`)
- [x] Fix array handling (`arr[]=xxx`)
- [x] Fix complex bitwise operation
- [x] Implements RAW mode (fix #826)
- [x] INI_SCANNER_TYPED


Indigo744:
Wait, the tests passed? It's not supposed to work! weird

jakubmisek:
@Indigo744 may I ask for write access to the `PARSE_INI` branch?

jakubmisek:
> Wait, the tests passed? It's not supposed to work!

The problem might be different PHP installations with different behavior on various test machines :)

jakubmisek:
I'm surprised so many INI features are not implemented in PeachPie.. 

add to "Fix array handling", this is implemented already in parsing URL queries (request handling and parse_str() func https://github.com/peachpiecompiler/peachpie/blob/09ff74af0eca7fce830c8012ea6a3a94d26cddbc/src/Peachpie.Runtime/Utilities/NameValueCollectionUtils.cs#L78) so we should reuse it

Indigo744:
For the two remaining, I'm not really sure how to do this, so maybe I'll let you look at it?

For array handling (arr[]=xxx), I guess you see where we should go.

For complex bitwise operation, actually the main errors are with the negative operator `~`:
```php
result3 = ~PHP_CONST_1 ^ PHP_CONST_4

// PHP_CONST_1 = 1 and PHP_CONST_4 = 4
// PHP = -6 (which is NOT(1) XOR 4 = -6)
// Peachpie = -1 (which is... I don't know!)
```

jakubmisek:
I can take a look on array handling.

The operators would have to be debugged, maybe wrong order of execution?

Indigo744:
This is coming along nicely, the gap with PHP is dramatically reduce regarding parse_ini! 😺

jakubmisek:
regarding the last issue - Fix complex bitwise operation - seems `PhpIniParser.Literal` does not handle `'^'` or `' '` tokens at all, so it consumes the entire `PHP_CONST_1 ^ PHP_CONST_4` as a single constant name

Indigo744:
> 
> 
> regarding the last issue - Fix complex bitwise operation - seems `PhpIniParser.Literal` does not handle `'^'` or `' '` tokens at all, so it consumes the entire `PHP_CONST_1 ^ PHP_CONST_4` as a single constant name

Interesting. But then I wonder how it is able to perform a bitwise on an inexistent weirdly named const?

jakubmisek:
> Interesting. But then I wonder how it is able to perform a bitwise on an inexistent weirdly named const?
I guess, the weirdly named constant is not found, so it is interpreted as a string value, which is converted to number `0`.

PR #828, opened 2020-08-31
PDO: Fix NULL bound value handling (in order to work with foreign keys)

Indigo744:
Hello,

This one was complicated, but I'm glad I managed to fix it. My initial error was with SQLSRV driver but I managed to reproduce it in SQLite.

When `NULL` was used as a bound value, updating a value that was a foreign key would throw an exception regarding a conflict with the foreign key. It's actually because the `NULL` value was converted to an empty string when bound.

This fix check for IsNull and use the proper bind param.


jakubmisek:
cool, thanks!

PR #829, opened 2020-08-31
Issue with array_pad using array as value

Indigo744:
It took me a lot of soul-searching to find this little one.

There is an issue when using `array_pad` with an array as value.
It seems that the array is copied by reference and mutating it will mutate all the others.

I do not have the solution yet, but I've added the test cases for `array_pad` (and `array_combine`, since I discovered it through this one and I saw that no test was available).


Indigo744:
Well it was easy after all. Please review, I fear `DeepCopy` is quite expansive? I don't know if there is another way...

jakubmisek:
awesome!

PR #833, opened 2020-09-01
Target .Net Standard 2.0 for Peachpie.AspNetCore.* projects

### LINKED ISSUE #832: Peachpie.AspNetCore.* targets .Net Standard 2.1
Hello,

**PeachPie itself targets .Net Standard 2.0** and it's a good thing, since it allows to be compiled to a wide variety of targets, including the .Net Framework.

However, **Peachpie.AspNetCore.Web** (and Peachpie.AspNetCore.MVC) **targets .Net Standard 2.1**, which I believe to target .Net Core 3.x branch and benefits from the performance improvement.

This is problematic, since a lot of project (legacy, but not so!) are stuck with ASPNet Core 2.1 which is needed when targeting the full .Net Framework.

Targeting .Net Standrard 2.0 would enable more people to integrate PeachPie in their existing code base. It would also put all PeachPie packages on the same compatibility level.

.Net Standard compatibility page: https://docs.microsoft.com/en-us/dotnet/standard/net-standard

__Would it be possible to target .Net Standard 2.0 for Peachpie.AspNetCore.* packages instead of 2.1?__
What would be needed in order to do so? I am more than wiling to work on a PR, should you agree 😉 

Thank you for your consideration and have a nice day!

Indigo744:
Alright so here we are. Tries to fix #832 

The changes are... small? I expected more issue. It compiles fine in .Net Standard 2.0, except 2 methods in `Peachpie.AspNetCore.Web\SynchronizedTextWriter.cs`.

I choose to enable both .Net Standard and use conditional compiler directive so former code is left untouched for 2.1.

However, since the changes are very small, I wonder it would not be easier to simply enable only .Net Standard 2.0?

That said, I have no idea how to test it. Any pointers?

### DIFF
--- a/src/Peachpie.AspNetCore.Web/SynchronizedTextWriter.cs
+++ b/src/Peachpie.AspNetCore.Web/SynchronizedTextWriter.cs
@@ -16,6 +16,11 @@ namespace Peachpie.AspNetCore.Web
     /// </summary>
     sealed class SynchronizedTextWriter : TextWriter
     {
+
+#if NETSTANDARD2_0
+        private readonly char[] writeChars;
+#endif
+
         HttpResponse HttpResponse { get; }
 
         public override Encoding Encoding { get; }
@@ -27,6 +32,9 @@ namespace Peachpie.AspNetCore.Web
         {
             HttpResponse = response ?? throw new ArgumentNullException(nameof(response));
             Encoding = encoding ?? throw new ArgumentNullException(nameof(encoding));
+#if NETSTANDARD2_0
+            writeChars = new char[1];
+#endif
         }
 
         const int UTF8MaxByteLength = 6;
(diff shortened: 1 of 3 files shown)


src/Peachpie.AspNetCore.Web/SynchronizedTextWriter.cs
```diff
@@ -81,13 +85,21 @@ public override void Write(char[] chars, int index, int count)
 
         public override void Write(char value)
         {
+#if NETSTANDARD2_0
+            var chars = new char[1] { value };
```
jakubmisek, line 89: I'd recommend to pre-allocate this into an instance field.

Indigo744: Alright, but is there any advantages here in keeping a difference between 2.0 and 2.1? Is `Span<char>` that great for a single char?

jakubmisek: 2.1 gets rid of most of the allocations, Span gets allocated on stack (no allocation on heap, no garbage collection), Encoding strings does not allocate either.

src/Peachpie.AspNetCore.Web/SynchronizedTextWriter.cs
```diff
@@ -16,6 +16,11 @@ namespace Peachpie.AspNetCore.Web
     /// </summary>
     sealed class SynchronizedTextWriter : TextWriter
     {
+
+#if NETSTANDARD2_0
+        private readonly char[] writeChars;
```
jakubmisek, line 21: maybe just
```c#
readonly char[] writeChars = new char[1];
```
so you don't have to alter the ctor at line 35

Indigo744: I tried that, but it's not possible to initialize a char array that way...

Indigo744: Ah sorry. I don't know why but when I tried that, I got an error. I must be going crazy... 😓

Indigo744:
Thank you!

PR #838, opened 2020-09-01
Fix DateInterval format (%a)

Indigo744:
Hello,

This little PR fixes the `%a` format used in DateInterval after a `date_diff()`.

The comments was already here, but not fully implemented (and the computation also, albeit not exactly correct) so I simply finished it.

### DIFF
--- a/src/Peachpie.Library/DateTime/DateInterval.cs
+++ b/src/Peachpie.Library/DateTime/DateInterval.cs
@@ -66,11 +66,11 @@ namespace Pchp.Library.DateTime
         /// <summary>
         /// Exposed interface to access the total number of days, see <see cref="_days"/>
         /// </summary>
-        public PhpValue days { 
-            get {
-                return _days != null
-                    ? PhpValue.FromClr(_days.Value)
-                    : PhpValue.False;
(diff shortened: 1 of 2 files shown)


Indigo744:
Wait, I identified another issue.

Indigo744:
Alright, should be good.

src/Peachpie.Library/DateTime/DateInterval.cs
```diff
@@ -217,7 +230,6 @@ private protected void Initialize(DateInfo ts, bool negative)
         private protected void Initialize(TimeSpan ts)
         {
             _span = ts.Duration(); // absolutize the range
-            days = (int)ts.TotalDays;
```
jakubmisek, line 220: just checking - is it correct to not initialize "days" property here?

Indigo744: Yes, because otherwise the following test would fail:
```php
    $interval = new \DateInterval('P2D');
    echo print_r($interval->format('%r%a'), true).PHP_EOL; // <-- should return (unknown)
```

PR #840, opened 2020-09-01
Fix strtotime

Indigo744:
strtotime fails with `strtotime("first monday of 2020-01")`

I only added the test cases for now, but I have no idea how to tackle this. Any pointers?

### DIFF
--- a/tests/datetime/strtotime.php
+++ b/tests/datetime/strtotime.php
@@ -1,0 +1,15 @@
+<?php
+namespace date\strtotime;
+
+function test() {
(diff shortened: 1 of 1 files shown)


jakubmisek:
I was checking the issue down to the actual time parser.

1. The tokens like "first", "monday", etc. are parsed at `Generators/StrToTime.lex` (generated into `Parsers\StrToTimeScanner.cs`). Seems the parser does not know the "of" token at all.

2. "2020-01" is also an unknown combination in the parser, it expects three numbers, not just two.

So it is possible to update the parser (StrToTime.lex) with the new "grammar". Technically it is just a lexer, so it's very simple pattern matching

PR #841, opened 2020-09-02
array_unique: fix NULL value handling

Indigo744:
The correct code was actually commented. PHP returns NULL when parameter is NULL.

### DIFF
--- a/src/Peachpie.Library/Arrays.cs
+++ b/src/Peachpie.Library/Arrays.cs
@@ -2456,9 +2456,8 @@ namespace Pchp.Library
         {
             if (array == null)
             {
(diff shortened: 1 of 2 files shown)


Indigo744:
Actually, there are similar code for a lot of other functions (array_diff_*, ...).

@jakubmisek please advised. I feel like we should log the PHP exception and return NULL to match PHP implementation instead of throwing a script-ending exception...

jakubmisek:
I've noticed, `array_unique` does not compare values correctly and also it boxes every value to `object` ... creating an issue https://github.com/peachpiecompiler/peachpie/issues/844

Issue #834, opened 2020-09-01
Add support for COM (Windows)


Indigo744:
[COM extension](https://www.php.net/manual/en/book.com.php) is actually not supported.

What would be the steps needed to bring it to PeachPie?

Of course it would work only on Windows, so I guess it should be made in a specific project (something like `Peachpie.Library.Com`)?

jakubmisek:
The steps would be to create a project similar to Peachpie.Library or Peachpie.Library.Graphics.

The core of the functionality is probably the php class `COM`, which would have magic methods `__call`, `__get`, etc. It's implementation would be using basically `Type.GetTypeFromProgID( ... )` and `Activator.CreateInstance` :)

PR #845, opened 2020-09-03
com_dotnet extension: basic implementation (barely working)

Indigo744:
See #834 

I started to work on a basic implementation on [extension `com_dotnet` ](https://www.php.net/manual/en/book.com.php)

**What works:**
- Creating [COM object](https://www.php.net/manual/en/class.com.php)
  - codepage / typelib not supported
- Calling methods with arguments _which returns a simple value type (not COM objects)_
- Getting / Setting properties _for value types (not COM objects)_
- [com_create_guid()](https://www.php.net/manual/en/function.com-create-guid.php)

**What is missing:**

Almost everything...! But the main work would revolve around implementing the [Variant class](https://www.php.net/manual/en/class.variant.php) so the COM class could inherit from it. Also this class would be used for handling complex COM object which is not working at all right now.

**What should not be supported:**

- Creating [dotnet object](https://www.php.net/manual/en/class.dotnet.php) as .Net framework 4.0 and later are not supported.

### DIFF
--- a/Peachpie.sln
+++ b/Peachpie.sln
@@ -82,6 +82,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Peachpie.NET.SdkTests", "sr
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Peachpie.App.Tests", "src\Tests\Peachpie.App.Tests\Peachpie.App.Tests.csproj", "{C8A0A533-BB62-4CB5-8861-06522DCF5EC2}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Peachpie.Library.ComDotNet", "src\Peachpie.Library.ComDotNet\Peachpie.Library.ComDotNet.csproj", "{B9DE827D-79F5-4066-80C5-7535EB66B443}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
--- a/src/Peachpie.Library.ComDotNet/Com.cs
+++ b/src/Peachpie.Library.ComDotNet/Com.cs
@@ -1,0 +1,119 @@
+﻿using System;
+using System.Linq;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using Pchp.Core;
+
+namespace Peachpie.Library.ComDotNet
+{
+    /// <summary>
+    /// COM object
+    /// <para>TODO : implements VARIANT and inherit from it, which would allow returning VARIANT object from __call and __get
+    /// and working with it, see test case com_dotnet/com.php</para>
+    /// </summary>
+    [PhpType(PhpTypeAttribute.InheritName), PhpExtension("com_dotnet")]
+    public class COM
+    {
+        #region Private members
+
+        object comObject = null;
+
+        const BindingFlags MemberAccessCom =
+            BindingFlags.Public | BindingFlags.Instance | BindingFlags.IgnoreCase;
+
+        #endregion
+
+        #region Construction
+
+        public COM(Context ctx, string module_name)
+        {
+            __construct(ctx, module_name);
+        }
+
+        [PhpFieldsOnlyCtor]
+        protected COM()
+        {
+        }
+
+        public virtual void __construct(Context ctx, string module_name)
+        {
+            var comType = Type.GetTypeFromProgID(module_name);
+            if (comType == null)
+                return;
+
+            comObject = Activator.CreateInstance(comType);
+        }
+
+        public virtual void __construct(Context ctx, string module_name, string server_name)
+        {
+            var comType = Type.GetTypeFromProgID(module_name, server_name);
+            if (comType == null)
+                return;
+
+            comObject = Activator.CreateInstance(comType);
+        }
+
+        public virtual void __construct(Context ctx, string module_name, string server_name, int code_page)
+        {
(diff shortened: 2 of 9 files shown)


Indigo744:
We should find a way to run the test only on Windows with specific extension activated? Not sure how to do this...

jakubmisek:
> We should find a way to run the test only on Windows with specific extension activated? Not sure how to do this...

for now, at the beginning you can do something like
```php
if (PHP_OS != "WINNT") exit("***SKIP***");
```

The test will be marked as skipped

Indigo744:
On the PeachPie side, should I add a test in the COM constructor (and in the function) where I test that we are on Windows and throw an `NotSupportedException` if not?

jakubmisek:
@Indigo744 I wouldn't check for Windows platform, what if it gets supported by .NET in the future (through magic) .. I guess .NET throws something anyways now

Indigo744:
Noted @jakubmisek 

However, I'm not sure I'll be able to push this implementation further in the short term.

Would it be acceptable to merge it "as is"? Some stuff are working, so it would be better than nothing for people needing this. How could we warn about what is missing?

jakubmisek:
I guess tests :) and also before the release, we are generating the compatibility matrix on https://docs.peachpie.io/compatibility-status/ which you can generate for yourself at https://github.com/peachpiecompiler/peachpie-docs/tree/master/tools

Indigo744:
> 
> 
> I guess tests :) and also before the release, we are generating the compatibility matrix on https://docs.peachpie.io/compatibility-status/ which you can generate for yourself at https://github.com/peachpiecompiler/peachpie-docs/tree/master/tools (it creates list of functions/classes in all loaded extensions - for PHP and PeachPie - and compares them)

I was able to generate a new updated list, but the com_dotnet extension sits at `0 / 107`... I am missing the relevant package on the PeachPie side, but I couldn't get it to work, even with the `1.0.0-dev` version..

jakubmisek:
you have to add a package/project reference into `generate.msbuildproj`

Indigo744:
@jakubmisek I was missing the lib from the local Nuget packages! Updating the script fixes this issue.

However, now I have the following error:
```
Unhandled exception. System.NotImplementedException: System.Reflection.BindingFlags
   at Pchp.Core.Dynamic.ConvertExpression.BindToValue(Expression expr)
   at Pchp.Core.Reflection.PhpPropertyInfo.ClrFieldProperty.CompileAccess(AccessMask access)
   at Pchp.Core.Reflection.PhpPropertyInfo.ClrFieldProperty.<.ctor>b__9_0()
   at System.Lazy`1.ViaFactory(LazyThreadSafetyMode mode)
   at System.Lazy`1.ExecutionAndPublication(LazyHelper executionAndPublication, Boolean useDefaultConstructor)
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.get_Value()
   at Pchp.Core.Reflection.PhpPropertyInfo.ClrFieldProperty.GetValue(Context ctx, Object instance)
   at Pchp.Library.Reflection.ReflectionClass.getConstants(Context ctx)
   at CallSite.Target(Closure , CallSite , PhpValue , Context )
   at System.Dynamic.UpdateDelegates.UpdateAndExecute2[T0,T1,TRet](CallSite site, T0 arg0, T1 arg1)
   at <Root>.program_php.collect(Context <ctx>) in peachpie-docs\tools\generate-compatibility\program.php:line 45
   at <Root>.program_php.<Main>(Context <ctx>, PhpArray <locals>, Object this, RuntimeTypeHandle <self>) in peachpie-docs\tools\generate-compatibility\program.php:line 65
   at <Script>.Main(String[] args)
```

I guess it doesn't know how to handle the `BindingFlags` type related to the constant `MemberAccessCom`...

jakubmisek:
@Indigo744 this is because PHP reflection tries to deal with the following constant:
```C#
const BindingFlags MemberAccessCom
```

which should not be visible to PHP code at all. Annotate the constant with `[PhpHidden]` attribute so it won't get into PHP runtime at all (or mark it as `internal protected`).

Indigo744:
`internal protected` was not enough, I still had the error. Adding `[PhpHidden]` did the trick however!

jakubmisek:
@Indigo744 sorry, `private protected` hides the field/const/method from PHP reflection

Indigo744:
Do you want me to modify it further, or is it good enough for now?

jakubmisek:
> Do you want me to modify it further, or is it good enough for now?

the attribute is ok for now

jakubmisek:
Hi @Indigo744 - can I help with this somehow? :)

Indigo744:
Hey @jakubmisek! Sorry for not getting back. I am swamped in work right now 😞 so as you could see I wasn't able to contribute as much as I could.

However note that this PR _could_ be merged as is. Some stuff are working. The rest could be done later. As you wish 😸

jakubmisek:
@Indigo744 just checking :) would be great to have a list of feature that are not implemented so we would have an overview.

I'd be happy to implement whats missing eventually, and merge the PR

Indigo744:
I've outlined in the PR description what is defined/what is missing.

An exact list is available in the docs PR: https://github.com/peachpiecompiler/peachpie-docs/pull/5.

Indigo744:
@jakubmisek Hello! Wow can't believe it's already 3 years later. Time flies!

Should we try merging it? I can clean up/fix conflicts.

jakubmisek:
> @jakubmisek Hello! Wow can't believe it's already 3 years later. Time flies!
> 
> Should we try merging it? I can clean up/fix conflicts.

that would be awesome :)

PR #846, opened 2020-09-03
Avoid throwing ArgumentNullException when array is NULL

Indigo744:
Following #841 

Avoid throwing ArgumentNullException when array is NULL on array_* functions to match PHP behavior


jakubmisek:
thank you @Indigo744 !

PR #847, opened 2020-09-07
setcookie: fix when name or value are NULL

Indigo744:
In [setcookie](https://www.php.net/manual/en/function.setcookie.php):

PHP behavior 

- when name is NULL or empty: return FALSE and add warning "Cookie names must not be empty"
- when value is NULL : return TRUE

PeachPie behavior
- when name is NULL or empty: throw exception
- when value is NULL: throw exception

This PR aims to match PHP behavior.


jakubmisek:
thanks :)

PR #875, opened 2020-10-24
Fixed linebreaks support in wordwrap()

kripper:
Test case:
`die(wordwrap("ab a\nbbbb\nccccc\nddd\neeee", 5));`


jakubmisek:
@kripper  thanks for the PR! seems it's still one character off of what's getting PHP, see the second test case. Any idea how to fix that?

PR #1031, opened 2022-01-22
Feature/local debugging

kendallb:
Add support for local builds of .net 4.61. I need to use this to be able to build correctly and debug .net 4.61 versions of the library. I cannot seem to debug .net standard. Can we somehow get this into the master tree as well?

### DIFF
--- a/build/dummy-net/dummy-net.msbuildproj
+++ b/build/dummy-net/dummy-net.msbuildproj
@@ -1,0 +1,16 @@
+﻿<Project Sdk="Peachpie.NET.Sdk/1.0.0-dev">
+
+  <PropertyGroup>
+    <TargetFramework>net461</TargetFramework>
+  </PropertyGroup>
+  
+  <ItemGroup Condition=" '$(DisableImplicitPeachpieReferences)' != 'true' ">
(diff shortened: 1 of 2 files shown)


jakubmisek:
Isn't it sufficient to add 
```xml
<TargetFrameworks>netcoreapp3.1;net462</TargetFrameworks>
````

to the [dummy.msbuildproj](https://github.com/peachpiecompiler/peachpie/blob/738c59b23dd0e864f613ed1c49f7570ef8744b51/build/dummy/dummy.msbuildproj#L4) ?

although wouldn't it start failing on Linux/Mac?

kendallb:
its been a while since I wrote that so I could debug it. When you build it in master as local debugging, it works for you? I wonder if I am missing some dependencies or something as it won't link for me in our solution.

jakubmisek:
I'm using the `dummy` project with `netcoreapp3.1` and it works for me for .NET Framework as well; (it causes the NuGet cache update with all its targets)

kendallb:
Ok, deleting this. It all works now. I have no idea why I needed that ages ago when I first was debugging it, but it works fine now.

PR #1038, opened 2022-02-01
improve the performance of 'fread' for tcp socket

avriltank:
I found the .net recv becomes very slow when the buffer size is big,it mallocs memory for
the byte[] array,but the gc do not free the memory soon

### DIFF
--- a/src/Peachpie.Runtime/PhpString.cs
+++ b/src/Peachpie.Runtime/PhpString.cs
@@ -98,9 +98,11 @@ namespace Pchp.Core.Text
                 // TODO: other types
 
                 default:
-                    throw new NotSupportedException(value.TypeCode.ToString());
+                    throw InvalidValueException(value);
             }
         }
+
+        static Exception InvalidValueException(PhpValue value) => new NotSupportedException(value.TypeCode.ToString());
 
         /// <summary>
         /// Copies characters to a new array of <see cref="char"/>s.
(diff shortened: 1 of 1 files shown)


src/Peachpie.Library/FileSystem.cs
```diff
@@ -615,6 +621,39 @@ public static PhpString fread(Context ctx, PhpResource handle, int length)
...
 
+            //improve the performance of 'fread' for tcp socket
+            //I found the .net recv becomes very slow when the buffer size is big,it malloc memory for
+            //the byte[] array,but the gc dot not free the memory soon
+            if(stream is SocketStream  theStream)
+            {
+                if(RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+                {
+                    var theSocket = theStream.Socket;
+                    if((int)theSocket.ProtocolType==6)//Tcp
```
jakubmisek, line 632: there is an enum for TCP

jakubmisek:
we try to avoid "unsafe" code, isn't there an issue on dotnet for this? Can we just make the buffer smaller?

avriltank:
There is no issue,the Socket.Receive method is very slowly when the buffer size is big.

avriltank:
Yes,you are right!there is not a good idea for peachpie code,i try to find another solution.

jakubmisek:
> I test in my project,the unsafe method can improve Almost 2.5x performance

it is possible, but we can't implement it like this. 

I would recommend implementing a C# method somewhere in your project, for example

```c#
PhpString fread_fast(Context ctx, PhpResource handle, int length) { ... }
```

Or we would need to find a "cleaner" workaround, so we can test it and maintain it.

avriltank:
Yes,that is a good idea,thanks for your reply

PR #1066, opened 2022-09-07
Fix C14N

kripper:
@jakubmisek here is a patch for the C14N bug.
Can you also please merge my helper function "peachpie_set_string_encoding()" which is required to change encoding and be able to work with ISO-8859-1 sources.
It won't harm the project, but will be usefull/required to work with ISO-8859-1 sources.


jakubmisek:
thanks @kripper

src/Peachpie.Library/Strings.cs
```diff
@@ -37,6 +37,11 @@ internal static CharMap InitializeCharMap()
 
         #endregion
 
+        // Helper function to change Console.OutputEncoding from PHP
+        public static void peachpie_set_string_encoding(string str) {
```
jakubmisek, line 41: We can't just add a public function to the API like that. You can always define this method in privately your project, or we could add a configuration option for this.

kripper: > We can't just add a public function to the API like that. You can always define this method in privately your project, or we could add a configuration option for this.

The configuration options sounds good to me.

jakubmisek:
Can we have the PR without the change to `Strings.cs`, it's totally unrelated to the issue.

thank you

kripper:
Please note that the helper method needed for working with iso-8859-1 and utf-8 at the same time, was reverted here: https://github.com/peachpiecompiler/peachpie/commit/280800183060e500a1c0ae0020663ec1b7536970

We have been using this method in production for years and we really need it, since there is no other option.

Having this method in our code, is also an obstacle for us to release pull requests of all other fixes that are required upstream.

We don't know how to deal with this method.

PR #1175, opened 2026-06-02
Fix http headers

xavierjefferson:


xavierjefferson:
@dotnet-policy-service agree

jakubmisek:
Thank you @xavierjefferson !

